PR #64, opened 2025-01-15
Update cbor-edn package to 1.1.0

chrysn:
### Description

This updates the cbor-edn package, which is used as part of aiocoap demos that use pyodide and Jupyter. The new version supports processing the full range of the notation as it is being updated, and brings readability enhancements by converting tagged data into application oriented literals.

Before: `{1:54(h'20010db8000000000000000000000001'),3:1(1736861269)}`

After: `{1: IP'2001:db8::1', 3: DT'2025-01-14T13:27:49Z'}`

The test is updated to cover some new functionality; as in the original version, it is by far not comprehensive, and mainly checks whether it's there and does something (but now that something includes features from the 1.1 series).

### Context

Closes: https://github.com/pyodide/pyodide/pull/5326

Filing this here after being redirected in https://github.com/pyodide/pyodide/pull/5326 -- is there any change log to update, or is this auto-managed here?

### DIFF
--- a/packages/cbor-diag/meta.yaml
+++ b/packages/cbor-diag/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: cbor-diag
-  version: 1.0.1
+  version: 1.1.0
   top-level:
     - cbor_diag
 source:
-  url: https://files.pythonhosted.org/packages/94/87/29ce613f876cd0ceb99ed1db71f7e3d90460c1babd2133ce30f18d5f540d/cbor_diag-1.0.1.tar.gz
-  sha256: 77e25a27681d59b0074a07aa494da0ab71f0ea6a826a2fd28167048d08a7ebcc
+  url: https://files.pythonhosted.org/packages/db/f6/9921162053f195eaa7e22cae9172e0d3bdda62282d76480163d1de872f35/cbor_diag-1.1.0.tar.gz
+  sha256: 78a85ab1165c43d224dc6b93e1ee791aec2f392c55d219d11937f78974f4e6c3
 build:
   script: |
     cargo update -p proc-macro2 --precise 1.0.60
--- a/packages/cbor-diag/meta.yaml
+++ b/packages/cbor-diag/meta.yaml
@@ -6,9 +6,6 @@ package:
 source:
   url: https://files.pythonhosted.org/packages/db/f6/9921162053f195eaa7e22cae9172e0d3bdda62282d76480163d1de872f35/cbor_diag-1.1.0.tar.gz
   sha256: 78a85ab1165c43d224dc6b93e1ee791aec2f392c55d219d11937f78974f4e6c3
-build:
-  script: |
-    cargo update -p proc-macro2 --precise 1.0.60
 requirements:
   executable:
     - rustup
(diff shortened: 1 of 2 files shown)


ryanking13:
> is there any change log to update, or is this auto-managed here?

We don't have a way to manage changelogs for this repository yet. I'll be writing code to calculate the diff between the old and new versions of the package list and update the list automatically, so I don't think you'll need to update the changelog yourself.

chrysn:
Building has run into trouble around an old proc-macro pinning in the package. I'm not sure that removing the script is fine, but let's see what CI says; the commit message is phrased on the assumption that it does (for if it does not, it won't get merged anyway).

ryanking13:
Thanks. I triggered the CI (for first-time contributors). Let's see if it works.

chrysn:
As I couldn't trace down precisely why CI failed, I ran a `pip install pyodide_build` / `pyodide build-recipes cbor-diag`, but couldn't get the precise emscripten version locally.

Is there an easy way to read the content of the build-reqs file which

```
ERROR Command '['/home/runner/miniconda3/envs/pyodide-env/bin/python', '-m',    
'pip', '--python', '/tmp/build-env-p2hkpgdr/bin/python', 'install',             
'--use-pep517', '--no-warn-script-location', '--no-compile', '-r',              
'/tmp/build-reqs-iukvkvg3.txt']' returned non-zero exit status 1.               
[2025-01-16 07:18:17] Failed building package cbor-diag in 14.0 seconds.
```

tripped over?

ryanking13:
build-reqs file will contains the [build-system requirements](https://github.com/chrysn/cbor-diag-py/blob/9d3490c7c3f6a29610b656eab7a48cdcb265ea74/pyproject.toml#L1-L3) in the pyproject.toml file.

We are currently [pinning the maturin version < 1.8](https://github.com/pyodide/pyodide-recipes/blob/eb15b15360a8b11bdf510935120fa42e9be2311c/tools/constraints.txt#L7) becuase of the regression in maturin but it seems like cbor-diag uses maturin >= 1.8 which is causing the error I guess.

chrysn:
Thanks. Following the #87 approach would not have made a difference with the old cbor-diag version (https://github.com/pyodide/pyodide/pull/5283), but current cbor-diag already sets `dynamic = ["version"]`, so it should not need the blocker now.

I'll keep an eye out for https://github.com/pyodide/pyodide/pull/5283 and #87, and if either makes progress, I can rebase this.

chrysn:
Thanks for addressing this with #97.

AIU, that alone is insufficient (I'll need another CI unblock after the rebase to see): The hard constraint on maturin < 1.8 will need to be removed from \*waves hands\* the system, and the packages that break after that will need their explicit constraint added. Are there any proposed changes around already to which I could subscribe (and then retry this once they are through)?

ryanking13:
> Are there any proposed changes around already to which I could subscribe (and then retry this once they are through)?

@chrysn I'll let you know after I sync the changes in pyodide/pyodide into this repository. Right now, we are going to apply some big changes that will affect package builds a lot (https://github.com/pyodide/pyodide/pull/5320,  https://github.com/pyodide/pyodide/pull/5332).

PR #67, opened 2025-01-27
Sync recipe (20250228)

ryanking13:


ryanking13:
@rgbkrk We updated the complier and Python version in Pyodide, so those wheels need to be recompiled with the new ABI. We are tracking the status for those packages in #99. Feel free to contact the maintainers and help them updating the wheels.

PR #69, opened 2025-02-06
Fix renamed import of MetaConfig

bartbroere:
It seems like the `MetaConfig` class has been moved from `pyodide_build.io` to `pyodide_build.recipe.spec`. It caused CI in a different draft PR to fail.


ryanking13 (approved):
Right, Thanks!

PR #70, opened 2025-02-08
Update lakers-python to 0.5.0

chrysn:
The new version supports a wider range of credentials, and will be a dependency of the upcoming release of aiocoap, which uses Jupyter notebooks as part of its introduction materials.


ryanking13:
Thanks! I applied packages updates in https://github.com/pyodide/pyodide-recipes/pull/67, so closing this as resolved.

PR #89, opened 2025-05-09
Try to re-enable CasADi

agriyakhetarpal:


ryanking13:
```
wasm-ld: error: unable to find library -ldlmalloc                               
wasm-ld: error: unable to find library -lcompiler_rt-legacysjlj                 
wasm-ld: error: unable to find library -lc++-legacyexcept                       
wasm-ld: error: unable to find library -lc++abi-legacyexcept                    
wasm-ld: error: unable to find library -lunwind-legacyexcept          
```

I think you need to add

```
embuilder build libdlmalloc libcompiler_rt-legacysjlj libc++-legacyexcept libc++abi-legacyexcept libunwind-legacyexcept --pic
```

to make sure these libraries are built.

https://github.com/pyodide/pyodide-recipes/pull/79/commits/fa12b7c53f2bcafea554ff228d5c1eca34638e65

agriyakhetarpal:
Thanks! I think we had dropped libraries from the xbuildenv quite some time ago. Would it make sense to add some of them, like these low-level libraries, back? What do you think?

agriyakhetarpal:
I see, thanks for the information!

agriyakhetarpal:
For `em++: error: undefined exported symbol: "_AddIdentity" [-Wundefined] [-Werror]`, the AddIdentity symbol comes from CasADi's internal source for SUNDIALS. It's defined in SUNDIALS in these places: https://github.com/search?q=repo:LLNL/sundials+addidentity&type=code

However, CasADi uses quite an old version of SUNDIALS, while the upstream library is now at 7.3.0.

I wonder if patching the function definition with `EMSCRIPTEN_KEEPALIVE` will help, but otherwise I don't know how to debut this further. I think it stopped working with Emscripten 3.1.66 or some version when we disabled it back then.

agriyakhetarpal:
Thanks, @hoodmane! I have to rely on using CI here, as the latest nightly build environment includes NumPy 2.0.2 and that fails to compile on my macOS machine (it tries to build from source because we're using 3.13) due to https://github.com/numpy/numpy/pull/27416

hoodmane:
Maybe we can manually backport that patch? If not maybe I can try debugging. Also what about using the nightly xbuildenv?

agriyakhetarpal:
Specifically, I notice three things:
- the fact that `whole_archive` is working here for `exports:`, but `requested` is not
- the [`AddIdentity` symbol](https://github.com/casadi/casadi/blob/a1e0fd19794f535c3e57b5ef7d1da252541f548b/external_packages/casadi-sundials/sundials/sundials_direct.c#L255-L271) isn't found because the linker tries to look for `_AddIdentity` with a leading underscore
- my patch to add `EMSCRIPTEN_KEEPALIVE` to `SUNDIALS_EXPORT` did not help and wasn't required, so perhaps I need to add it to only the function definitions.

and these have something to do with each other, and I need to figure out what that is, beyond saying, "it works with `--whole-archive` because it doesn't mangle/add an underscore before this symbol".

Also, yes, SUNDIALS is statically linked as far as I can see from the artifacts in the build logs: `[ 40%] Linking C static library ../../lib/libcasadi_sundials.a` and in the code as `add_library(casadi_sundials STATIC ${SUNDIALS_INCLUDES} ${SUNDIALS_SOURCES})`

However, this is how SUNDIALS_EXPORT is defined in `casadi/external_packages/casadi-sundials/sundials/sundials_config.h`:

```h
/* Mark SUNDIALS API functions for export/import
 * When building shared SUNDIALS libraries under Windows, use
 *      #define SUNDIALS_EXPORT __declspec(dllexport)
 * When linking to shared SUNDIALS libraries under Windows, use
 *      #define SUNDIALS_EXPORT __declspec(dllimport)
 * In all other cases (other platforms or static libraries under
 * Windows), the SUNDIALS_EXPORT macro is empty
 */
#define SUNDIALS_EXPORT
```

and given that the recipe worked in Emscripten 3.1.58 and hasn't worked since Emscripten 3.1.64 (or maybe it was broken with 3.1.59, that could be possible).

agriyakhetarpal:
I'll now be trying a combination of `exports: requested` + adding `EMSCRIPTEN_KEEPALIVE` to the `AddIdentity` function. If it passes for AddIdentity but fails somewhere else during linkage, that would suggest that we need to mark `-s EXPORTED_FUNCTIONS` to all symbols marked with `SUNDIALS_EXPORT`, if I am understanding things correctly.

ryanking13:
> I, too, think we should be with whole_archive, but I do want to get to the bottom of the issue, as I think a patch that works should be upstreamable.

Sure, understanding the root cause and upstreaming the fix would be the best. Thanks!

hoodmane:
> the [AddIdentity symbol](https://github.com/casadi/casadi/blob/a1e0fd19794f535c3e57b5ef7d1da252541f548b/external_packages/casadi-sundials/sundials/sundials_direct.c#L255-L271) isn't found because the linker tries to look for _AddIdentity with a leading underscore

The underscore prefixing is just a confusing thing that Emscripten does when talking about symbols. I don't think that's likely to be a true symptom of the problem.

PR #90, opened 2025-05-11
Add `rustworkx` alpha recipe

IvanIsCoding:
Related to https://github.com/Qiskit/rustworkx/pull/1447

In short, this adds a very experimental `rustworkx` to Pyodide. If this is the wrong place to submit a recipe, let me know.

At the current status, the other `rustworkx` author has not yet merged the Pyodide support. It also might take a while for us to publish 0.17.0. So I called it a pre-release with `0.17.0a3` and hosted it in my GitHub fork's release. 

Overall, this seems to be working on the browser but failing with Python from `pyodide venv` so I will open an issue on `pyodide-build` asking a couple questions. (edit: report in https://github.com/pyodide/pyodide-build/issues/201)

### DIFF
--- a/packages/rustworkx/meta.yaml
+++ b/packages/rustworkx/meta.yaml
@@ -9,7 +9,7 @@ source:
   url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
   sha256: 8bd0c295134e2b0c03808d4e69428b41153849db6488839084a78793c337f191
 build:
-  script: export RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT"
+  script: export RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT -Z emscripten-wasm-eh"
 requirements:
   executable:
     - rustup
--- a/packages/rustworkx/meta.yaml
+++ b/packages/rustworkx/meta.yaml
@@ -13,6 +13,9 @@ build:
 requirements:
   executable:
     - rustup
+test:
+  imports:
+    - rustworkx
 about:
   home: https://github.com/Qiskit/rustworkx
   PyPI: https://pypi.org/project/rustworkx
--- a/packages/rustworkx/meta.yaml
+++ b/packages/rustworkx/meta.yaml
@@ -9,7 +9,7 @@ source:
   url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
   sha256: 8bd0c295134e2b0c03808d4e69428b41153849db6488839084a78793c337f191
 build:
-  script: export RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT -Z emscripten-wasm-eh"
+  script: export RUSTFLAGS="$RUSTFLAGS -C target-feature=+atomics,+bulk-memory,+mutable-globals"
 requirements:
   executable:
     - rustup
--- a/packages/rustworkx/meta.yaml
+++ b/packages/rustworkx/meta.yaml
@@ -1,0 +1,23 @@
+package:
+  name: rustworkx
+  version: 0.17.0a3
+  top-level:
+    - rustworkx
+  tag:
+    - rust
+source:
+  url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
+  sha256: 8bd0c295134e2b0c03808d4e69428b41153849db6488839084a78793c337f191
+build:
+  script: export RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-s -C link-arg=ALLOW_MEMORY_GROWTH=1 -C link-arg=-s -C link-arg=MODULARIZE=1 -C link-arg=-s -C link-arg=EXPORT_NAME=createModule -C link-arg=-s -C link-arg=EXPORT_ES6=1 -C link-arg=-s -C link-arg=ASSERTIONS=1"
+requirements:
+  executable:
+    - rustup
+about:
+  home: https://github.com/Qiskit/rustworkx
+  PyPI: https://pypi.org/project/rustworkx
+  summary: A python graph library implemented in Rust
+  license: Apache 2.0
+extra:
+  recipe-maintainers:
+    - IvanIsCoding
--- a/packages/rustworkx/meta.yaml
+++ b/packages/rustworkx/meta.yaml
@@ -20,7 +20,7 @@ about:
   home: https://github.com/Qiskit/rustworkx
   PyPI: https://pypi.org/project/rustworkx
   summary: A python graph library implemented in Rust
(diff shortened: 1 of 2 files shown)


packages/rustworkx/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+  version: 0.17.0a3
+  top-level:
+    - rustworkx
+  tag:
+    - rust
+source:
+  url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
+  sha256: 8bd0c295134e2b0c03808d4e69428b41153849db6488839084a78793c337f191
+build:
+  script: export RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-s -C link-arg=ALLOW_MEMORY_GROWTH=1 -C link-arg=-s -C link-arg=MODULARIZE=1 -C link-arg=-s -C link-arg=EXPORT_NAME=createModule -C link-arg=-s -C link-arg=EXPORT_ES6=1 -C link-arg=-s -C link-arg=ASSERTIONS=1"
```
hoodmane, line 12: I think you should get rid of this `RUSTFLAGS` setting. It surely won't with this.
```suggestion
```
If you do need to add something to `RUSTFLAGS` try to do it like this:
```
   export RUSTFLAGS="$RUSTFLAGS -C <some new flag>"
```
The stuff that is in there by default is important.

IvanIsCoding: The reality is the build just need s`"-C", "target-feature=+atomics,+bulk-memory,+mutable-globals"` in addition to the default flags. But for `pyodide build` setting `RUSTFLAGS` doesn't seem to be additive?

hoodmane: You can do `export RUSTFLAGS="$RUSTFLAGS -Ctarget-feature=+atomics,+bulk-memory,+mutable-globals`. But `bulk-memory` and `mutable-globals` are on by default. Atomics probably isn't.

hoodmane:
> Package Build Results

Wow that's fancy @ryanking13

IvanIsCoding:
Ok, I reused the `RUSTFLAGS`. It will be pretty evident if the script does not capture the defaults, the tests will fail I think.

For the other target features, I didn't get a very conclusive answer by running `rustc +nightly -Z unstable-options --print target-features --target=wasm32-unknown-unknown`. It just listed all available features. Atomic for sure is an unstable option, but I am not 100% confident the other two are enabled by default. So I will leave them even if they are somewhat redundant

hoodmane (approved):
Looks good to me if tests pass.

packages/rustworkx/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: rustworkx
+  version: 0.17.0a3
+  top-level:
+    - rustworkx
+  tag:
+    - rust
+source:
+  url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
```
agriyakhetarpal, line 9: Non-blocking; could we perhaps use the PyPI sdist here?

packages/rustworkx/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+  executable:
+    - rustup
+test:
+  imports:
+    - rustworkx
+about:
+  home: https://github.com/Qiskit/rustworkx
+  PyPI: https://pypi.org/project/rustworkx
+  summary: A python graph library implemented in Rust
+  license: Apache 2.0
```
agriyakhetarpal, line 23:
```suggestion
  license: Apache-2.0
```
nit: please change this to the SPDX short identifier

agriyakhetarpal (approved):
Thanks, @IvanIsCoding ! The tests are passing, and I had two comments, this should be good to merge after them. I'm glad you were able to figure out the out-of-tree build, too.

packages/rustworkx/meta.yaml
IvanIsCoding: I'd like to eventually, but the current status is:
* 0.16 will fail to build as it does not contain https://github.com/Qiskit/rustworkx/pull/1445
* https://github.com/Qiskit/rustworkx/pull/1447 is not yet reviewed nor merged
* We've never published a pre-release on PyPI, I think our release setup could work but  that is TBD. So assume 0.17 will take a couple of weeks to release

So publishing a "patched" alpha release to my fork seemed like a fair compromise. The moment I get a PyPI release I'll send a PR updating

agriyakhetarpal: Sounds good to me, thanks for the info!

IvanIsCoding:
Ok, I will be able to add a check on my CI. I think I found a way to make things work with a pinned Pyodide Build 0.30.2, Pyodide 0.27.5, and Emscripten 3.1.58 via https://pixi.sh/.

Soon that Pyodide version will be outdated, but I think as long as a build with 0.27.5 keeps passing, the tests in `pyodide-recipes` will catch any trivial errors with newer versrirons.

agriyakhetarpal:
Sounds good to have a smoke test, but I'm also a little confused about your PRs – why not run the full test suite in the CI using `pyodide venv` and installing the WASM wheel within it? We don't run the test suite here as the recipes are what have smoke tests, so that we can get to distributing/updating packages faster, plus we unvendor the tests using rudimentary mechanisms to reduce wheel size. However, in your own CI, there is no such limitation and you could run the tests on every commit? This is what many scientific Python packages are doing – they either build the wheel using `pyodide build` + run the tests by installing the WASM wheel, pytest, etc. in a Pyodide virtual environment, or they use `cibuildwheel`'s `--platform pyodide` option which does the same thing with a neater interface.

agriyakhetarpal:
BTW, the Python interpreter provided by `pyodide venv` is also just some code running with Node.js as a part of a polyglot binary – it should be more Pythonic to use, and you can avoid adding the JS script this way. :)

agriyakhetarpal:
Another thing I noticed from the pixi setup – while it mentions that it is possible only with with linux-x86-64; I'd like to note that linux-arm64, macos-amd64, and macos-arm64 should all work equally well. We test for macOS in pyodide-build's CI.

IvanIsCoding:
> Sounds good to have a smoke test, but I'm also a little confused about your PRs – why not run the full test suite in the CI using `pyodide venv` and installing the WASM wheel within it? We don't run the test suite here as the recipes are what have smoke tests, so that we can get to distributing/updating packages faster, plus we unvendor the tests using rudimentary mechanisms to reduce wheel size. However, in your own CI, there is no such limitation and you could run the tests on every commit? This is what many scientific Python packages are doing – they either build the wheel using `pyodide build` + run the tests by installing the WASM wheel, pytest, etc. in a Pyodide virtual environment, or they use `cibuildwheel`'s `--platform pyodide` option which does the same thing with a neater interface.

We can move to somewhere as this PR is approved but for the full test suite: I think `rustworkx` will be able to run a full test soon. We use `unittest` so to be honest I think I might be able to just mount our test code and write something very basic `unittest.TestLoader` and `unittest.TextTestRunner`.

Why not use `pyodide venv`?  I wanted to have the least dependencies and commands possible.
. I pinned on a Pyodide version from npm, Node is pretty stable and it seems like pyodide does a good job with its own lockfile.

And for the Linux part in Pixi: you opened https://github.com/conda-forge/emscripten-feedstock/issues/35. If it got closed I'd add macOS support for aarch64. I'd also add Linux support for aarch64 but they don't have that either. And Windows users

IvanIsCoding:
https://github.com/Qiskit/rustworkx/pull/1450 now has the full test suite. Fortunately, all tests pass. The only ones that are skipped are the ones that required graphviz. Honestly, the result is better than anticipated.

agriyakhetarpal:
I don't think there has been anything else to do here, so I guess we can merge now!

PR #91, opened 2025-05-16
Fix `python-flint` build failure [full build]

agriyakhetarpal:
For some reason `python-flint` failed on `main`: https://github.com/pyodide/pyodide-recipes/actions/runs/15062463362/job/42340019530

In pyodide/pyodide#5626, we pinned Cython<3.1, but this seems to be resolved as python-flint 0.7.1 (which is being updated to here) is now compatible with Cython 3.1.0a1.

This PR updates the versions of `python-flint` and FLINT, and updates FLINT's build script for WASM compilation.

PR #92, opened 2025-05-16
Try out parallel testing [full build]

agriyakhetarpal:
In this PR, I'm trying to investigate if we can shorten the time taken to run the test suite using `pytest-xdist`, which can take upwards of ~25 minutes in the case of a full build.


agriyakhetarpal:
Of course, we can only get the best estimate using a full build, but if the tests all work correctly and pass in this parallel context, it should be a sign that there shouldn't be any (many?) problems.

agriyakhetarpal:
Based on https://github.com/pyodide/pyodide-recipes/actions/runs/15079325254/job/42394053702?pr=92, this seems to work! 409 tests executed in 1 minute and 51 seconds, while in [the logs for an older job](https://github.com/pyodide/pyodide-recipes/actions/runs/14987623595/job/42138842177?pr=90#step:11:1505), I see that 380 tests ran in 3 minutes. The time savings will show up better and will be more useful for full builds.

hoodmane:
Previously we had xdist on but our ci was pretty flakey and @rth turned it off because he thought it was oversubscribed. Because the browser uses several threads. Maybe start with xdist on node?

agriyakhetarpal:
Oh, I didn't know! I'll see if I can restrict this to just Node through a global-level marker and let the rest of the tests, but I don't immediately see it in the matrix. I thought that we were testing only with Chrome at first, but that doesn't seem to be the case. I'll check.

agriyakhetarpal:
Okay, I notice that we aren't running tests on Node.js in CI at the moment, it's just that the `conftest.py` file was ported and includes some references to be used later. I assume there has been a decent amount of time since the last leg of `pytest-xdist` usage in the main Pyodide repo. I could trigger the tests a few more times to see if there are any flakes, if you want me to, but we could also go ahead and revert the change in case we see there is a problem that isn't easily fixable. What do you think?

Thanks!

hoodmane:
I'm fine with merging it and seeing what happens. Maybe don't set the concurrency very high? 

We should also add node tests ideally.

agriyakhetarpal:
> I'm fine with merging it and seeing what happens. Maybe don't set the concurrency very high?

Great! Yes, since this is on GitHub Actions runners, and we've set it to `-n auto`, so the number of jobs is 2 (and 3 on macOS, if we were to add it later). We would have more cores on a larger runner shall we opt for one.

agriyakhetarpal:
> We should also add node tests ideally.

Yes, and also Safari and Firefox, as we just need to download the artifacts in a matrix of jobs. I can take it up in a follow-up.

agriyakhetarpal:
Let's wait for #91 before we go ahead with this; I'll rebase after that is done.

ryanking13:
Thanks. Yeah, I think it was flaky before, but it was a long time ago, so we can try it again. We can revert it anytime if it doesn't work.

agriyakhetarpal:
> > -n auto, so the number of jobs is 2
> 
> Just worth double checking that this is creating the number of jobs you expect. "auto" may not work properly in some virtualized environnement including CI [pytest-dev/pytest-xdist#1103](https://github.com/pytest-dev/pytest-xdist/issues/1103). as they are not handling some of the edge cases that say [joblib.cpu_count](https://joblib.readthedocs.io/en/latest/generated/joblib.cpu_count.html#joblib.cpu_count) includes...

Thanks for that link, @rth. I tried to find the CPU specifications for the Linux runners, but I don't see any, unfortunately. If these are Intel runners, then we need to check for hyper-threading, and the CPU affinity might be a better value to use. If they are AMD CPUs, we have SMT which isn't well defined either as we know nothing about the make/model/architecture.

It looks like there are now 4 cores across all runners, bumped up from the previous 2 (except M-series macOS runners, which have 3): https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories. The page doesn't say anything about the number of physical/logical/virtual cores. In the meantime, we can change to `-n ${nproc}`, hopefully that helps.

> Thanks. Yeah, I think it was flaky before, but it was a long time ago, so we can try it again. We can revert it anytime if it doesn't work.

agriyakhetarpal:
> > -n auto, so the number of jobs is 2
> 
> Just worth double checking that this is creating the number of jobs you expect. "auto" may not work properly in some virtualized environnement including CI [pytest-dev/pytest-xdist#1103](https://github.com/pytest-dev/pytest-xdist/issues/1103). as they are not handling some of the edge cases that say [joblib.cpu_count](https://joblib.readthedocs.io/en/latest/generated/joblib.cpu_count.html#joblib.cpu_count) includes...

Thanks for that link, @rth. I tried to find the CPU specifications for the Linux runners, but I don't see any, unfortunately. If these are Intel runners, then we need to check for hyper-threading, and the CPU affinity might be a better value to use. If they are AMD CPUs, we have SMT which isn't well defined either as we know nothing about the make/model/architecture.

It looks like there are now 4 cores across all runners, bumped up from the previous 2 (except M-series macOS runners, which have 3): https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories. The page doesn't say anything about the number of physical/logical/virtual cores. In the meantime, we can change to `-n ${nproc}`, hopefully that helps.

agriyakhetarpal:
I've triggered a full build and full test suite run, let's see if we have any issues.

ryanking13:
Thanks! The failing tests are related to cython so I think they are okay.

The output looks a little bit noisy than before (two lines per test?). It would be nice to fix that, but it is not a big problem, otherwise looks good to me.

ryanking13 (approved):

agriyakhetarpal:
Yes, I'm not really sure why that is happening. I think we can try again; I'm happy to wait a little. It could be a bug with how pytest-xdist interacts with the terminal and reports test successes/failures.

agriyakhetarpal:
Based on multiple outputs in the logs, here's an example:

```
[gw2] [ 10%] PASSED packages/blosc2/test_blosc2.py::test_large_typesize[chrome-shape0-261-False] 
packages/blosc2/test_blosc2.py::test_large_typesize[chrome-shape0-65536-True]
```

What I think is happening here is that it uses multiple lines because it adds the `[ 10%] PASSED` text to the first test case out of a parametrised test, but it doesn't do so for the rest of the parametrised test cases that come after it?

agriyakhetarpal:
Okay, after a bit of experimentation, I think this is expected. I ran the `pyodide-build` test suite, and this is what we get on a few of the tests:

```bash
[gw0] [ 59%] PASSED pyodide_build/tests/test_venv.py::test_venv_cli_args[options2-expected_calls2]
pyodide_build/tests/test_venv.py::test_venv_cli_args[options4-expected_calls4]
pyodide_build/tests/test_venv.py::test_venv_cli_args[options5-expected_calls5]
pyodide_build/tests/test_venv.py::test_supported_virtualenv_options
pyodide_build/tests/test_venv.py::test_venv_creation[default]
```

This means that there is something more fishy going on, and it's not a problem with this repository alone – I checked the logs for other tests and this is everywhere. Perhaps https://github.com/pytest-dev/pytest-xdist/issues/605?

A worker $X$ that operates on a test method $Y$ with $n$ parametrized cases first runs the starting one, and then it proceeds to display the next case out of the $(n - 1)$ cases it's going to run within the next lines. However, I confirmed that no tests are missing from the test report.

Hence we should be fine to merge this, as it's easily a close-to-2x improvement (31 minutes ➡️ 16 minutes). 

Thanks for the reviews!

PR #97, opened 2025-05-21
Enable packages iminunit, ppl, and primecount

ryanking13:
- iminunit, ppl: Just works, probably LLVM issue in Emscripten 4.0.7.
- pyxel: manually modified build directory
- primecount: updated the package to fix clang version issue.

### DIFF
--- a/packages/iminuit/meta.yaml
+++ b/packages/iminuit/meta.yaml
@@ -1,8 +1,6 @@
 package:
   name: iminuit
   version: 2.30.1
-  # Doesn't build on Emscripten 4.0.7
-  _disabled: true
 source:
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
Yes,  I hope `iminuit` and `primecount` will work here 🤞🏻 Maybe we can seek some advice from @kitao here for the `pyxel` build issue? :)

ryanking13:
Maybe, yeah, but I think it is more like a Pyodide side issue. I noticed that it has something to do with https://github.com/emscripten-core/emscripten/issues/24106.

ryanking13:
Okay... I think I find a way to fix pyxel. Let me open a separate PR.

PR #105, opened 2025-06-06
Fix Clarabel.rs

ryanking13:
Faer < 0.22.0 has issue with 32-bit platforms.

https://github.com/sarah-quinones/faer-rs/issues/173


agriyakhetarpal (approved):
Looks good to me, thanks, @ryanking13!

PR #108, opened 2025-06-13
Update Narwhals to 1.42.1

MarcoGorelli:


ryanking13:
Looks like there is a failure in the unittest. Could you take a look?

PR #111, opened 2025-06-16
Update httpx

hoodmane:
This fixes a few bugs:
* the timeout could sometimes be None when a float was expected
* the async client didn't handle content-encoding correctly
* I merged https://github.com/encode/httpx/pull/3207 to make errors pickleable

I also refactored stuff to extract common code between the async
and sync code paths as much as I could.

I added some tests here.


hoodmane:
We need to install `pyodide-py`.

PR #112, opened 2025-06-17
chore: Update recipe for css-inline

Stranger6667:
This PR updates the `css-inline` recipe

Ref: https://github.com/Stranger6667/css-inline/issues/447


agriyakhetarpal:
Thank you for working through this, @Stranger6667!

PR #115, opened 2025-06-19
Add soundfile package and required libraries

swnf:
### Description

Based on pyodide/pyodide#5154

This PR adds the [soundfile](https://github.com/bastibe/python-soundfile) python package and its required libraries. In more detail, this PR
 - Compiles libflac, libopus, and libmp3lame as static libraries
 - Compiles libsndfile as a static library (I could not convince its build system to use the `SIDE_MODULE_LDFLAGS` to link a shared library).
 - Patches soundfile to use cffi's API mode instead of the ABI mode (I don't think ABI mode can work in pyodide without more changes)
 - Patches soundfile to remove its custom wheel tagging logic


packages/libmp3lame/meta.yaml
```diff
@@ -0,0 +1,22 @@
+package:
+  name: libmp3lame
+  version: "3.100"
+  tag:
+    - library
+    - static_library
+source:
+  url: https://sourceforge.net/projects/lame/files/lame/3.100/lame-3.100.tar.gz
```
ryanking13, line 8: Sourceforge sometimes blocks the request from CI, so could you please host it somewhere else? For instance, you can make a GitHub repository for hosting the file, and upload the file there.

ryanking13 (approved):
Thanks. I left a minor comment about the download URL, otherwise looks good to me.

packages/libmp3lame/meta.yaml
swnf: I've replaced the URL with VLC's download server

PR #117, opened 2025-06-21
Cryptography 45.0.4

greateggsgreg:
Depends on fix for this: https://github.com/pyodide/pyodide/issues/5709


ryanking13 (approved):
This would not fix the regression, but I think it is okay to merge. Thanks!

PR #120, opened 2025-06-23
ujson 5.10.0 - New package

greateggsgreg:
Added ujson package, one of the fastest drop-in replacements for the built-in Python json module

### DIFF
--- a/packages/ujson/meta.yaml
+++ b/packages/ujson/meta.yaml
@@ -1,0 +1,129 @@
+package:
+  name: ujson
+  version: 5.10.0
+  top-level:
+    - ujson
+source:
+  url: https://files.pythonhosted.org/packages/source/u/ujson/ujson-5.10.0.tar.gz
+  sha256: b3cd8f3c5d8c7738257f1018880444f7b7d9b66232c64649f562d7ba86ad4bc1
+about:
+  home: https://github.com/ultrajson/ultrajson
+  PyPI: https://pypi.org/project/ujson
+  summary: Ultra fast JSON encoder and decoder for Python
+  license: "Developed by ESN, an Electronic Arts Inc. studio.
+Copyright (c) 2014, Electronic Arts Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+* Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


packages/ujson/meta.yaml
```diff
@@ -0,0 +1,129 @@
...
+  top-level:
+    - ujson
+source:
+  url: https://files.pythonhosted.org/packages/f0/00/3110fd566786bfa542adb7932d62035e0c0ef662a8ff6544b6643b3d6fd7/ujson-5.10.0.tar.gz
+  sha256: b3cd8f3c5d8c7738257f1018880444f7b7d9b66232c64649f562d7ba86ad4bc1
+about:
+  home: https://github.com/ultrajson/ultrajson
+  PyPI: https://pypi.org/project/ujson
+  summary: Ultra fast JSON encoder and decoder for Python
+  license: "Developed by ESN, an Electronic Arts Inc. studio.
```
ryanking13, line 13: Could you summarize the license with a shorter sentence? I know that this is auto-generated field, but it is a bit too long. The license field here doesn't have to be a full license.

packages/ujson/meta.yaml
```diff
@@ -0,0 +1,129 @@
+package:
+  name: ujson
+  version: 5.10.0
+  top-level:
+    - ujson
+source:
+  url: https://files.pythonhosted.org/packages/f0/00/3110fd566786bfa542adb7932d62035e0c0ef662a8ff6544b6643b3d6fd7/ujson-5.10.0.tar.gz
```
agriyakhetarpal, line 7: With predictable URLs from PyPI, this can be simplified as follows:

```suggestion
  url: https://files.pythonhosted.org/packages/source/u/ujson/ujson-5.10.0.tar.gz
```

agriyakhetarpal (commented):
Thanks @greateggsgreg! This looks great to me, and I left two suggestions below. Additionally, I couldn't help but wonder if we should add `ujson` to the distribution, considering that https://github.com/ultrajson/ultrajson#project-status suggests migrating to `orjson` instead, which we already distribute. Though I suppose that the notice shouldn't stop us from distributing `ujson`.

packages/ujson/meta.yaml
greateggsgreg: Done based on @agriyakhetarpal ’s suggestion. Thank you!

PR #121, opened 2025-06-23
Pyskein 1.0 - New package

greateggsgreg:


agriyakhetarpal (changes requested):
This package appears not to have been updated in twelve years, as indicated by https://pypi.org/project/pyskein/#history. Do you happen to have a maintained fork or a different package in mind that implements the Skein algorithm, @greateggsgreg?

greateggsgreg:
> This package appears not to have been updated in twelve years, as indicated by https://pypi.org/project/pyskein/#history. Do you happen to have a maintained fork or a different package in mind that implements the Skein algorithm, @greateggsgreg?

Unfortunately it's a pretty obscure hashing algorithm that lost the bid to be the next implementation in the SHA3 standards competition. OpenSSL implementation is unlikely because of this, and this is the only implementation that supports all three block sizes of the Skein hashing algorithm. The only other Python implementation I could find was this: https://pypi.org/project/geesefly/ which does not support all block sizes.

In the time it was in the SHA3 competition, there were a few projects that used it to include a file protection app, a cryptocurrency coin, and other programs that chose to implement exotic hashing algorithms.'

The purpose of my recent PRs is in support of porting my corporate app to run in Pyodide and ensure that our dependencies will build properly in the hopes that they will all continue to work when the Emscripten Python PEP is merged. This is certainly an obscure dependency so we're happy to handle this out-of-tree while it's needed if there isn't general benefit.

agriyakhetarpal:
Thanks for the context, @greateggsgreg! I suppose it isn't too bad to add here, but I think it could require a non-trivial amount of porting changes to the code, packaging infrastructure, etc. to support Python 3 (considering that [v1.0 _dropped support_ for Python 3.0](https://pythonhosted.org/pyskein/download.html#changes-between-version), how often do we see that 😅) and adapt to modern Python packaging standards. If you could set up a `pyskein-ng` or `pyskein2` package on PyPI with the aforementioned changes, and additionally add an appropriate license, vet its dependencies and code for any security concerns as feasible (as long as it's not malware – heh), add project-health-specific files such as a code of conduct and the like, then I think we could make a case to add it here. If it is a pure Python package and does not have any compiled components, it does not necessarily need to be added to the distribution – `await micropip.install("pyskein-ng")` would work well, and you could handle it out-of-tree as you mentioned.

ryanking13:
Yeah, I'm also a little reluctant to add a package that was last updated in 2013. Even if the Python code in this package is very small, I think it's hard to fix if something breaks.

PR #134, opened 2025-06-25
sourmash-4.9.3 - Bump version

ctb:


agriyakhetarpal:
I triggered the workflows, thanks!

> looks like I need workflow approval.
> 
> Let me know if there's a way you prefer for new packaging people to request such approval, etc. etc. and I can add something to the top-level README, too.

We have set the approval requirement in the settings as "Require approval for first-time contributors". We could add a team for recipe maintainers to this repository, perhaps, but it could be too intrusive to send out invite emails to so many members. To avoid this requirement for now, you could add yourself as a maintainer for `sourmash` in another PR, so that you would no longer be a first-time contributor?

agriyakhetarpal:
Yes, that would be great!

ryanking13:
Looks like there is some error:

```
ffi.error: symbol 'revindex_free' not found in library '/lib/python3.13/site-packages/sourmash/_lowlevel/lib_lowlevel.so': Tried to lookup unknown symbol "revindex_free" in dynamic lib: /lib/python3.13/site-packages/sourmash/_lowlevel/lib_lowlevel.so
```

Do you understand what the error means? I can take a look at the binary if you don't have any clue.

ctb:
Thanks! I have a strong suspicion as to what is going on, but to verify I will need to cut a new release of both the sourmash Rust crate _and_ the sourmash Python project, so it may take me a few days 😭 .

ctb:
> Getting RocksDB to work in WASM might be a non-trivial task. I see that it is not specified in the dependencies in `pyproject.toml` though, @ctb?

It's in the Rust layer, [src/core/Cargo.toml](https://github.com/sourmash-bio/sourmash/blob/latest/src/core/Cargo.toml#L110). And look there, it even says it doesn't compile under wasm 😭

We usually use Rust features to turn it off where we don't or can't use it, so I just need to figure out the right magic. And conveniently that will also let me poke some of the other features that don't work in wasm (https://github.com/sourmash-bio/sourmash/issues/3709)!

I think I'll close this for now. We're going to need another release of sourmash before this can work ;).

PR #138, opened 2025-06-27
Rename zlib to libzlib

hyoinandout:
Towards #142

### DIFF
--- a/packages/libhdf5/meta.yaml
+++ b/packages/libhdf5/meta.yaml
@@ -13,7 +13,7 @@ source:
 
 requirements:
   host:
-    - zlib
+    - libzlib
 
 build:
   type: shared_library
--- a/packages/libxml/meta.yaml
+++ b/packages/libxml/meta.yaml
@@ -10,7 +10,7 @@ source:
 
 requirements:
   host:
-    - zlib
+    - libzlib
     - libiconv
 
 build:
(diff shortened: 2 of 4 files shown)


hyoinandout:
Hi, thank you for the review.

As mentioned in [the previous PR](https://github.com/pyodide/pyodide/pull/5703), I will seek for more packages that would be better to be renamed and push commits to this PR. Somehow, since you already approved this PR, I can merge it right away, before this PR being deprecated. Do you have any preference regarding this manner?

agriyakhetarpal:
Hi @hyoinandout, both approaches are acceptable here, though I would suggest merging this PR as it is ready. At the time you opened pyodide/pyodide#5703, the recipes had not been moved to this repository, and building all of them took a considerable amount of time in CI. The benefit will be that the workflows will automatically run for you when you create a PR to rename more packages, and we won't need to approve and run them – we only require approval from first-time contributors.

agriyakhetarpal (approved):
Thanks, @hyoinandout!

agriyakhetarpal:
Perhaps I merged too early: https://github.com/pyodide/pyodide-recipes/actions/runs/15964421692/job/45022197659, we should have run the full builds here first 😅 I'll take the responsibility of fixing it, sorry about that!

PR #140, opened 2025-06-29
Add `pyodide-build` repository as a submodule

agriyakhetarpal:
## Description

This PR adds `pyodide-build` as a submodule, just like it is the case in the Pyodide main/runtime repository. This has a few advantages:
- being able to test out changes to `pyodide-build` to assess breakages for new changes or compatibility with older changes
- being able to modify compilation routines such as `pywasmcross.py` to assess changes
- being able to test out ABI changes with new nightly xbuildenvs
- and so on

As we've pinned our Pyodide build environment in `pyproject.toml` in the repository root, this PR does not introduce any ABI or compilation-specific changes.

> [!NOTE]
> The note below has been resolved; the submodule points to https://github.com/pyodide/pyodide-build/commit/ee11882923d8e06e8164324fec5d225e4db3bbaf on the `main` branch now.

> [!TIP]
> ~The submodule checked in this PR builds upon my branch, which I've submitted as a PR at https://github.com/pyodide/pyodide-build/pull/213. Once merged, I shall update the submodule to point to `main`. Please review that PR first. Thank you!~

## PR stack

Please review the following PRs in the order listed below:

- https://github.com/pyodide/pyodide-build/pull/168
- https://github.com/pyodide/pyodide-build/pull/213
- https://github.com/pyodide/pyodide-recipes/pull/140 ➡️ **you are here**
- https://github.com/pyodide/pyodide-recipes/pull/141

PR #141, opened 2025-06-29
Update to SciPy v1.17.1, and update SciPy test suite infrastructure to use `pyodide venv`

agriyakhetarpal:
## Description

This PR bumps SciPy to version 1.17.1, released on February 23, 2026: https://github.com/scipy/scipy/releases/tag/v1.17.1. Here are a few points:

- I partially reverted https://github.com/scipy/scipy/pull/24407 to restore support for using `numpy.f2py` to call the custom target generators for remaining Fortran extension modules
- One patch has been dropped, so we're down from twelve to eleven now
- Previously, the test suite used to be stuck at the 14-15% range, but with our SciPy v1.17.0 update, we had noticed that this was no longer happening. I'm happy to say it doesn't happen with v1.17.1 either :D

<details><summary>Old description, kept for archival purposes, from a time when this PR was trying to update to SciPy v1.16.0:</summary>
<p>

- The number of patches has been reduced; we now have fourteen patches from the previous eighteen (although there is always scope to add a few more once this is implemented).
	- The patch previously applied for `id_dist` has been dropped, and a patch for ARPACK based on the C port https://github.com/scipy/scipy/pull/22748 has been introduced.
	- A lot of other troublesome FORTRAN77 code has been removed since v1.14.1 notably: MINPACK, QUADPACK, L-BFGS-B, SLSQP, NNLS, `linalg.expm`, `id_dist`, and other Fortran codes that are not listed here; please view https://github.com/scipy/scipy/issues/18566 for more.
- The problem of not being able to use the `f2py` entry point directly via Meson generators remains, and we have to resort to using` python -m numpy.f2py` and custom targets until pyodide/pyodide-build#21 is resolved.
- Many new C-rewritten packages also have the same problem of [alternate returns](https://docs.oracle.com/cd/E19957-01/805-4940/6j4m1u7qv/index.html), which means that we need to patch their signatures to match the ones coming from OpenBLAS, which means `void`s are transformed to `int`s in the code for the ARPACK, the `_matfuncs` family, FITPACK, SLSQP, NNLS, and the L-BFGS-B extension modules.
- Most of the diff of 50K lines added here comes from the ARPACK code; and the patch is more or less the same as the aforementioned upstream PR with hardly any differences. The removal of 20K lines comes from dropping the `id_dist` patch,
- Unfortunately, I tried testing the previous tests that we used to run via `scipy-pytest.js`, but the tests always hang on my macOS machine locally at the ~14-15% stage at various points, despite skipping many of them. I suggest that we leave the extra testing for another PR, as the necessary infrastructure to get those tests up and running is a follow-up task that I'd be happy to assign myself to.

> [!TIP]
~This PR depends on https://github.com/pyodide/pyodide-recipes#140, which, at the time of writing, contains a patch that allows compilation to proceed to completion and succeed. That PR also depends on another PR (pyodide/pyodide-build#168).~ Please review those PRs first (starting with the latter) before coming back to this one, as this will reduce the diff to the relevant files. Thank you!

## PR stack

Please review the following PRs in the order listed below:

- https://github.com/pyodide/pyodide-build/pull/168
- https://github.com/pyodide/pyodide-build/pull/213
- https://github.com/pyodide/pyodide-recipes/pull/140
- https://github.com/pyodide/pyodide-recipes/pull/141 ➡️ **you are here**

## Additional context

SciPy was previously updated in https://github.com/pyodide/pyodide/pull/5031 to version 1.14.1.

Follow-up tasks after this PR:
- Sync this update to Pyodide, and address https://github.com/pyodide/pyodide/pull/5715
- Remove `unvendor-tests: true` and use `-Cinstall-args=--tags=runtime,python-runtime,devel`
- Start working on https://github.com/pyodide/pyodide/issues/5268
- Port https://github.com/pyodide/pyodide/pull/5060

</p>
</details>

<hr>

Tagging the recipe maintainers and our previous collaborators here, like I've done in the past: @lesteve, @steppi, and @rgommers

Thank you for your time!


ryanking13 (approved):

rgommers:
> Hi @rgommers, I experimented a bit with the f2py generator definition code to make `find_program()` use `python -m numpy.f2py` instead of just `f2py`, but I didn't get very far with this – I thought that would be good to get done with to drop one more patch (patch 0014) here to get to thirteen patches in total.

If it's working now for Pyodide, I'd prefer to not touch this at all right now and keeping the patch, because we're close to proper cross-compilation support in meson-python and then we should be able to get it right at once.

rgommers:
> Many new C-rewritten packages also have the same problem of [alternate returns](https://docs.oracle.com/cd/E19957-01/805-4940/6j4m1u7qv/index.html),

Can you elaborate? The link says that alternate returns are a Fortran feature with no equivalent in C.

rgommers:
Okay thanks, that is clearer. So it's not that the new C code in SciPy is problematic, but that the OpenBLAS signatures have already been changed and hence the corresponding SciPy code that calls into OpenBLAS needs changing as well. 

It seems to me that that's a problem inherent in using LAPACK interfaces in Pyodide, and the only way around it is to use LAPACKE instead. Might be worth revisiting at some point after we've finished removing all Fortran code in SciPy.

agriyakhetarpal:
> Okay thanks, that is clearer. So it's not that the new C code in SciPy is problematic, but that the OpenBLAS signatures have already been changed and hence the corresponding SciPy code that calls into OpenBLAS needs changing as well.

Yes, indeed!

> It seems to me that that's a problem inherent in using LAPACK interfaces in Pyodide, and the only way around it is to use LAPACKE instead. Might be worth revisiting at some point after we've finished removing all Fortran code in SciPy.

I think so too. Until we can explore LAPACKE, LLVM Flang from v20 and up is our best bet as a direct Fortran-to-WASM compiler; pyodide/pyodide#5268 has a few details about previous (partially successful/not yet successful) efforts.

agriyakhetarpal:
Okay, the full test suite is back now – I'm going to trigger it and see what new test failures show up. I hope there won't be too many :)

agriyakhetarpal:
Yes, unfortunately, https://github.com/pyodide/pyodide-recipes/actions/runs/15982819048/job/45082284069?pr=141 has been stuck at this test for the past six minutes (at the time of writing):

```
tests/test_differentiate.py::TestDerivative::test_accuracy[numpy-case51]
```

which I can also replicate locally. We're running the `"not slow"` tests. The CI job hangs at 3%, but at least I can get to 15% on my end. I wonder if using a `pyodide venv` will be of any help to us here. I think it will be quite challenging to determine the reason for such hangs, as I am unsure if there are any methods at our disposal to access any debugging information.

agriyakhetarpal:
All the errors in the test suite are coming from `_dierckx.cpython-313-wasm32-emscripten.so`, so I guess it's now time to perform some surgery on it and get some diagnostics.

agriyakhetarpal:
```
wasm-objdump /Users/agriyakhetarpal/Desktop/pyodide-recipes/packages/scipy/.venv-pyodide/lib/python3.13/site-packages/scipy/interpolate/_dierckx.cpython-313-wasm32-emscripten.so -j Export -x
```

suggests that the symbol does exist:

```
_dierckx.cpython-313-wasm32-emscripten.so:      file format wasm 0x1

Section Details:

Export[4]:
 - func[39] <__wasm_call_ctors> -> "__wasm_call_ctors"
 - func[42] <PyInit__dierckx> -> "PyInit__dierckx"
 - func[59] <_ZNSt3__212__tuple_implINS_15__tuple_indicesIJLm0ELm1EEEEJNS_6vectorIdNS_9allocatorIdEEEENS3_IxNS4_IxEEEEEEC2B8ne190106IJLm0ELm1EEJS6_S8_ETpTnmJEJEJRS6_RS8_EEENS1_IJXspT_EEEENS_13__tuple_typesIJDpT0_EEENS1_IJXspT1_EEEENSE_IJDpT2_EEEDpOT3_> -> "_ZNSt3__212__tuple_implINS_15__tuple_indicesIJLm0ELm1EEEEJNS_6vectorIdNS_9allocatorIdEEEENS3_IxNS4_IxEEEEEEC2B8ne190106IJLm0ELm1EEJS6_S8_ETpTnmJEJEJRS6_RS8_EEENS1_IJXspT_EEEENS_13__tuple_typesIJDpT0_EEENS1_IJXspT1_EEEENSE_IJDpT2_EEEDpOT3_"
 - func[40] <__wasm_apply_data_relocs> -> "__wasm_apply_data_relocs"
 ```

and from the `Type` section dump:

```
type[2] () -> nil
type[4] () -> i32
```

And from the `Function` section:

```
func[42] sig=4 <PyInit__dierckx>
```

So `PyInit__dierckx` has signature `type[4]` which is `() -> i32`. This is correct.

Perhaps I missed patching a `DLARTG` somewhere?

**Comment by ryanking13** [comment]
Thanks for the investigation, @agriyakhetarpal. The error happens while loading the library,

```
Failed to load lib  /Users/agriyakhetarpal/Desktop/pyodide-recipes/packages/scipy/.venv-pyodide/lib/python3.13/site-packages/scipy/interpolate/_dierckx.cpython-313-wasm32-emscripten.so
Error: Failed to load dynamic library /Users/agriyakhetarpal/Desktop/pyodide-recipes/packages/scipy/.venv-pyodide/lib/python3.13/site-packages/scipy/interpolate/_dierckx.cpython-313-wasm32-emscripten.so: 0
```

Because of the error during the loading step, the symbols will not be filled out in the GOT table. That's why the 

```
ImportError: dynamic module does not define module export function (PyInit__dierckx)
```

error happens.

To see a more helpful error log, try a debug build of Pyodide, which is available by passing the following flag `PYODIDE_SYMBOLS=1` when building the Pyodide runtime. It will show a much richer message in the library loading process.

**Comment by agriyakhetarpal** [comment]
Thanks, @ryanking13! I wonder if we could also release an `xbuildenv-debug.tar.bz2` in the same GitHub nightly/scheduled release, so that we won't have to compile it ourselves? We could also expand to `debugoptimized`, `debug`, and `release` builds (but only document the released ones for now, like we've been doing).

**Comment by agriyakhetarpal** [comment]
> > Failed to load lib  /Users/agriyakhetarpal/Desktop/pyodide-recipes/packages/scipy/.venv-pyodide/lib/python3.13/site-packages/scipy/interpolate/_dierckx.cpython-313-wasm32-emscripten.so
> > Error: Failed to load dynamic library /Users/agriyakhetarpal/Desktop/pyodide-recipes/packages/scipy/.venv-pyodide/lib/python3.13/site-packages/scipy/interpolate/_dierckx.cpython-313-wasm32-emscripten.so: 0
> 
> But I think this error only happens in the `pyodide venv` environment, so the hang in the unittest might have a different reason.

Oh, actually, another test in TestBSplines (when I was running the test suite locally) was hanging for the same reason (when I skipped this test to allow the test suite to proceed), so I think it must be related to `_dierckx` – it goes through the same code paths. So the `scipy-pytest.js` script with `node` also likely has the same error with loading this lib, but it's not being reported in stdout/stderr and the script is not crashing/failing here when it should.

> > Thanks, @ryanking13! I wonder if we could also release an xbuildenv-debug.tar.bz2 in the same GitHub nightly/scheduled release, so that we won't have to compile it ourselves? We could also expand to debugoptimized, debug, and release builds (but only document the released ones for now, like we've been doing).
> 
> Sounds good to me. Would you like to update [pyodide/pyodide-build-environment-nightly](https://github.com/pyodide/pyodide-build-environment-nightly?rgh-link-date=2025-07-01T08%3A21%3A11Z) to support that feature?

Yes, I'll put together a PR!

**Comment by ryanking13** [comment]
Meanwhile, I ran the hanging test locally and was able to reproduce it in my browser:

```
from scipy import stats
from scipy.differentiate import derivative
from js import console

distcont = stats._distr_params.distcont
for idx, case in enumerate(distcont):
	distname, params = case
	console.log("handling case: ", idx, distname, params)
	dist = getattr(stats, distname)(*params)
	x = dist.median() + 0.1
	res = derivative(dist.cdf, x)
	ref = dist.pdf(x)
```

The `derivative` function (which was introduced in scipy 1.15) hangs when it the case is

```
['jf_skew_t', (8, 4)],
```

**Comment by agriyakhetarpal** [comment]
Thanks! Yes, there might be a few more broken libraries besides `_dierckx`, as I had at least 6-7 test cases that were hanging. However, I didn't check if they were all going through `_dierckx`, as it is the only one being reported right now. 

Perhaps we could test individual test modules (`scipy.stats`, `scipy.io`, `scipy.special`, etc.) in the test suite to isolate this issue more effectively.

**Comment by agriyakhetarpal** [comment]
Also, may I ask how you are testing SciPy in your browser? When I copy the Pyodide runtime using the provided script and copy the OpenBLAS libs and other libs from the `.libs/` directory into `dist/`, and run Pyodide through a basic `test.html` file in the `dist/` directory, it complains about not being able to find the OpenBLAS lib in the directory. Subsequently, it fails to load the `_dop` module. I'm sure that I'm doing something wrong there!

**Comment by ryanking13** [comment]
> Also, may I ask how you are testing SciPy in your browser? When I copy the Pyodide runtime using the provided script and copy the OpenBLAS libs and other libs from the .libs/ directory into dist/, and run Pyodide through a basic test.html file in the dist/ directory, it complains about not being able to find the OpenBLAS lib in the directory. Subsequently, it fails to load the _dop module (much earlier than attempting to load _dierckx). I'm sure that I'm doing something wrong there!

For `pyodide venv` I think it is because the nightly buildenv does not have corresponding packages in the CDN. When running Pyodide in node, we dynamically download the package from the CDN when the package is not available in the filesystem using the version number in the Pyodide runtime, but for nightly buildenv, the version number would be something like `0.28.0dev0` so it probably fails to download packages like openblas.

**Comment by agriyakhetarpal** [comment]
Yes, it hangs at `test_differentiate.py::TestDerivative::test_accuracy[numpy-case51]`, still. We don't use `pyodide venv` here, though, just a JS script with Node.js. I'll switch to `pyodide venv` as it will report errors early, at least, instead of hanging.

**Comment by agriyakhetarpal** [comment]
@ryanking13 thanks for handling the NumPy version conflicts for Pyodide 0.29. I think this was inadvertently closed due to the "fix #141" string in the PR description 😅 Reopening it.

**Comment by ryanking13** [comment]
@ilayn  That's really good news! Thanks for sharing. Is there a planned timeline for the scipy 1.17 release?

**Comment by ianthomas23** [comment]
On `emscripten-forge` we've finally managed to build recent `scipy` using the `flang` Fortran compiler so it is possible, albeit with significant patching of `flang`, `openblas` and `scipy`. Packages for `scipy` `1.16.x` and `1.17.0` are hosted at [prefix.dev](https://prefix.dev/channels/emscripten-forge-dev/packages/scipy) and there is a demo `xeus-python` JupyterLite deployment using `scipy 1.17.0` at https://ianthomas23.github.io/scipy-3x-deploy/lab/index.html.

**Comment by agriyakhetarpal** [comment]
@ryanking13 thanks for including my older changes from here into pyodide/pyodide#5995 and pyodide/pyodide#6095. I've taken some time to update to v1.17.1, and fortunately, things are working as usual here. Context: I decided it'd be nice to spend time on this sooner rather than later, because the first release candidate for SciPy v1.18 is just around the corner: https://discuss.scientific-python.org/t/proposed-release-schedule-for-scipy-1-18-0/2308. I'll start out with that (and leave it open, to continue iterating on until the final release) after Pyodide v314.0.0 stable release duties :)

File: .github/workflows/build.yml [hunk 1]
```diff
@@ -196,6 +196,7 @@ jobs:
             # No changed recipes
             if [ "$SCIPY_WILL_BE_TESTED" == "true" ]; then
               echo "recipes=tag:core,scipy" >> "$GITHUB_OUTPUT"
+              sed -i 's/unvendor-tests: true/unvendor-tests: false/' packages/scipy/meta.yaml
```
ryanking13, line 199: Please double check that the tests are not bundled into the scipy wheel in production build.

ryanking13 (approved):
Thanks for your works on Scientific python support @agriyakhetarpal!

.github/workflows/build.yml
agriyakhetarpal: Yes, I think we should be fine. The only code that sets `unvendor-tests: false` is the `sed` command here, and it's gated by `if: github.event_name == 'pull_request'`. So for the release and release nightly jobs, we do a full build, and it doesn't change the `unvendor-tests` setting, and so the wheel uploaded to Anaconda always has `unvendor-tests: true`.

One thing I did catch after you prompted me to read the workflow files again (thanks!) was that the `workflow_dispatch:` event with `scipy=true` would run the SciPy test suite, but it would fail because it wouldn't retain the tests. I fixed that in 12e420cd8676596615b826a5e8a43666ac175919.

agriyakhetarpal:
Okay, all green, merging!

PR #143, opened 2025-06-30
Fix builds for `libnetcdf` and `lxml` after #138

agriyakhetarpal:
This PR fixes the build job after `zlib` was renamed to `libzlib` in #138, as these two packages were missed as a part of the changed recipes calculation.


hyoinandout:
Thank you:D

PR #148, opened 2025-06-30
Add SciPy test suite

agriyakhetarpal:


agriyakhetarpal:
Okay, the test suite is passing with the current SciPy v1.14.1 – this is as expected, as these tests were running in https://github.com/pyodide/pyodide/pull/5031. I'll suggest that we merge this for now, and explore any follow-ups here in subsequent PRs. Thank you!

PR #152, opened 2025-07-01
Speed up disk space cleanup

### LINKED ISSUE #151: "Clear disk space" step can be sped up
This step takes an awfully long time to complete; in many cases, it takes just about the same time or longer than it takes to build a recipe itself (if the recipe doesn't depend on many packages, that is). Maybe there is an action that can do this for us in a faster manner...

agriyakhetarpal:
Closes #151


ryanking13:
Nice!

PR #164, opened 2025-07-12
tests: import awkward-cpp Python binding

henryiii:
This is broken (https://github.com/scikit-hep/awkward/issues/3572), so seeing if we can expose the issue. Will try updating (44 is old) next.

Also, the awkward (pure Python) version must work with the awkward-cpp version, so maybe it makes sense to ship that too so it resolves correctly.

### DIFF
--- a/packages/awkward-cpp/meta.yaml
+++ b/packages/awkward-cpp/meta.yaml
@@ -11,7 +11,7 @@ source:
 build:
   exports: requested
   backend-flags: |
-    cmake.define.CMAKE_BUILD_WITH_INSTALL_RPATH=ON
+    cmake.define.CMAKE_MODULE_LINKER_FLAGS=-Wl,-rpath=\$ORIGIN
     build.verbose=ON
 
 requirements:


henryiii:
Still getting `ImportError: dynamic module does not define module export function (PyInit__ext)`.

ryanking13:
Thanks for the report. The problem is that the `awkward_cpp/lib/_ext.cpython-313-wasm32-emscripten.so` relies on `awkward_cpp/lib/libawkward.so`, but the RPATH is not set for `_ext.cpython-313-wasm32-emscripten.so` so it failes to locate the libawkward.so. Could you try passing `-rpath=$ORIGIN` when linking _ext.cpython-313-wasm32-emscripten.so?

ryanking13:
Meanwhile, we will probably need to update our import test to fail when loading the shared library fails.

ryanking13:
> Why is this different on 0.28, out of curiosity? (It also works on all other platforms)

Before 0.28, we hacked the Emscripten's library loading mechanism to locate the shared libraries. From 0.28, Emscripten officially supports RPATH (https://github.com/WebAssembly/tool-conventions/issues/244), so we are using it instead.

> And it does fail, the problem was it wasn't loading the extension, lib._ext isn't automatically imported when importing the outer package; it's not a user facing package (awkward is).

But the `import awkward` will import `_ext` module internally isn't it? So if the `_ext` module is not loaded correctly, `import awkward` will fail anyway.

ryanking13:
> This is already there:

Interesting, then probably Pyodide's build system is ignoring that part for some reason.

ryanking13:
I can't see `-rpath` passed to the compiler when building the `_ext.so`.

```
emcc:WARNING: invocation: /home/siha/dev/pyodide-recipes/emsdk/upstream/emscripten/em++.py -fPIC -O3 -DNDEBUG -fexceptions -flto -shared -o _ext.cpython-313-x86_64-linux-gnu.so CMakeFiles/_ext.dir/src/python/_ext.cpp.o CMakeFiles/_ext.dir/src/python/content.cpp.o CMakeFiles/_ext.dir/src/python/forth.cpp.o CMakeFiles/_ext.dir/src/python/io.cpp.o libawkward.so -Werror=implicit-function-declaration -Werror=mismatched-parameter-types -Werror=return-type -Wl,--fatal-warnings -O2 -g0 -L/home/siha/.cache/.pyodide-xbuildenv-0.30.6.dev5+gee11882/https_github_com_pyodide_pyodide-build-environment-nightly_releases_download_20250626_xbuildenv_tar_bz2/xbuildenv/pyodide-root/cpython/installs/python-3.13.2/lib/ -s WASM_BIGINT -fwasm-exceptions -sSUPPORT_LONGJMP -s SIDE_MODULE=1 -sSIDE_MODULE=2 -sEXPORTED_FUNCTIONS=@/tmp/tmp4e8nifdq (in /home/siha/dev/pyodide-recipes/packages/awkward-cpp/build/awkward-cpp-44/build/cpython-313)
```

---

```
- When the RPATH needs to be modified during installation (different from build-time RPATH), CMake may use binary patching tools
- Linux (ELF format): Uses tools like chrpath or patchelf to modify the ELF headers
- macOS (Mach-O format): Uses install_name_tool to modify the Mach-O load commands
```

When I asked LLM, it answered `CMAKE_INSTALL_RPATH` patches the binary directly to add RPATH instead of passing the `-rpath` flag in the linking step, which probably doesn't work because the WASM module format is totally different from the ELF binary. I am not a CMAKE expert, so I am not sure it is the correct answer, though. But if it is correct, probably adding [`BUILD_WITH_INSTALL_PATH`](https://cmake.org/cmake/help/latest/prop_tgt/BUILD_WITH_INSTALL_RPATH.html#prop_tgt:BUILD_WITH_INSTALL_RPATH) work?

henryiii:
Is there a way to set config-settings and/or environment variables in the recipes?

ryanking13:
> 
Is there a way to set config-settings and/or environment variables in the recipes?

You can add `build.script` to pass env variable

```yaml
build:
  script: |
    export ENV_VAR=...
```

or use `build.backend-flags` to pass config-settings to the build backend

```
build:
  backend-flags: |
    setup-args=-Dallow-noblas=true
```

henryiii:
Working on this in https://github.com/scikit-hep/awkward/pull/3575.

packages/awkward-cpp/meta.yaml
```diff
@@ -1,15 +1,18 @@
...
 source:
-  url: https://files.pythonhosted.org/packages/a0/9c/d7655969d075f99edb4d3d0e57ea63d4742af3bfff4953ed54c7224dfca5/awkward_cpp-44.tar.gz
-  sha256: 8dc499288d6d16b2ea20b51a27d5047e51a247b6aacfcbcb3b302cad6d3c87d8
+  url: https://files.pythonhosted.org/packages/78/e3/5a255643d0ffdd1b2c8e53c4e0376a073f0547b85daa18f20cefa14bd90f/awkward_cpp-47.tar.gz
+  sha256: 676cf4976810edab32187edf5a8a716af95047b9038c96d27d3be44f1331950f
 
 build:
   exports: requested
+  backend-flags: |
+    cmake.define.CMAKE_MODULE_LINKER_FLAGS=-Wl,-rpath=\$ORIGIN
```
henryiii, line 14: This is a little hacky but it works. I don't see a way to force CMake to add it otherwise. We can add this in the next version of Awkward, as well.

ryanking13 (approved):
Thanks!

cc: @agoose77 as a recipe maintainer for visibility.

PR #168, opened 2025-07-14
Support `document.pyodideMplTarget` in Matplotlib WebAgg backend

ianthomas23:
This PR adds support for use of `document.pyodideMplTarget` in the Matplotlib WebAgg backend to specify the parent element within which to append Matplotlib plots, falling back to `document.body` if it is not specified. This was a feature of the `matplotlib-pyodide` backends that I did not originally add to the patched WebAgg backend.

Usage is like this
```js
document.pyodideMplTarget = document.getElementById('target')
```
to insert plots in an element such as
```html
<div id='target'></div>
```

There is no explicit test for this, but I have tested it manually on my dev system.

There is a question of where documentation for such things should go in the future given that we probably want to avoid any package-specific information in the pyodide docs. Probably it should be in the Matplotlib docs under the [WebAgg backend](https://matplotlib.org/stable/users/explain/figure/backends.html#interactive-backends).

Fixes pyodide/pyodide#5751.

### DIFF
--- a/packages/matplotlib/test_matplotlib_backend.py
+++ b/packages/matplotlib/test_matplotlib_backend.py
@@ -470,3 +470,43 @@ def test_transparency(selenium_standalone):
 
     save_canvas_data(selenium, ref)
     compare_canvas_data(selenium, ref.read_bytes())
+
+
+@pytest.mark.parametrize("use_mpl_target", [False, True])
+@matplotlib_test_decorator
+def test_pyodide_mpl_target(selenium_standalone, use_mpl_target):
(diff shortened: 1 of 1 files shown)


hoodmane:
Thanks for the very fast response @ianthomas23! Code looks good to me. Should we add a test?

ianthomas23:
I've added a test, by querying the DOM in the `@run_in_pyodide` function.

ryanking13 (approved):
Thanks @ianthomas23!

PR #173, opened 2025-07-18
Add cobs

whitequark:


agriyakhetarpal (approved):
Thanks, @whitequark!

whitequark:
Is the package immediately available for use, or is there a release process involved?

agriyakhetarpal:
It will be available with the next release, which will include a new lockfile with this package that you can pass to `loadPyodide()`. I can trigger a release for you; it just needs a new `0.28-20250719` tag.

whitequark:
That would be super handy, thanks!

PR #174, opened 2025-07-21
Pymupdf2

linuxrider:


ryanking13 (commented):
Thanks, but it seems like there is an error locating `libmupdfcpp.so`.

Also ccing @sanzoghenzo who were the original author of the PR

ryanking13:
In Pyodide 0.28.0, we updated our package dependency resolution logic to use RPATH (https://github.com/emscripten-core/emscripten/issues/22126).

So, if the shared libraries in the wheel does not set the RPATH correctly, it may fail to locate the dependencies.

linuxrider:
I took the the build artifact from the CI and loaded it in the console.html.
Strangely, the error message is different every time. Finally, no error is displayed. Importing pymupdf fails, however.
image

linuxrider:
Thank you for your comments.

PR #303, opened 2025-07-30
Bump numpy 2.3.2

greateggsgreg:


ryanking13 (changes requested):
The numpy version needs to be synced with the Pyodide main branch. Let's pause this for a while, as we have some changes related to numpy #141

PR #312, opened 2025-08-05
Add fastapi

hoodmane:


ryanking13 (approved):
It looks like we need to disable some test in Firefox.

```
RuntimeError: WebAssembly stack switching not supported in this JavaScript runtime
```

Otherwise looks good to me. Thanks!

PR #314, opened 2025-08-08
Updated python-sat to version 1.8.dev19.

alexeyignatiev:
The update includes
- a few performance improvements to the existing tools dealing with over-constrained systems (like RC2, LBX, Hitman, and OptUx),
- a pair of brand-new tools for compiling arbitrary Boolean formulas into Blake Canonical Form (Primer) and computing the smallest size CNF and DNF representation of arbitrary Boolean formulas (Bica),
- a few bug fixes in arbitrary Boolean formula creation and "clausification".

### DIFF
--- a/packages/python-sat/meta.yaml
+++ b/packages/python-sat/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: python-sat
-  version: 1.8.dev17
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks!

PR #316, opened 2025-08-13
Bump rustworkx to 0.17.1 [full build]

IvanIsCoding:
Follow up to #90 

rustworkx now supports Pyodide in a released version! Previously, it used an alpha version from my personal repository.

Thanks again for providing `pyodide-build` and all the recipes!


IvanIsCoding:
I need to investigate the compilation failure, my apologies

IvanIsCoding:
There was nothing wrong with the library code per se, but right now `rustworkx` needs `rust-src` because of `-C target-feature=+atomics,+bulk-memory,+mutable-globals"` in the `RUSTFLAGS`.

A small change to `script` fixed the compilation issue.

packages/rustworkx/meta.yaml
```diff
@@ -1,15 +1,17 @@
...
     - rust
 source:
-  url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
-  sha256: 8bd0c295134e2b0c03808d4e69428b41153849db6488839084a78793c337f191
+  url: https://github.com/Qiskit/rustworkx/archive/refs/tags/0.17.1.tar.gz
+  sha256: dc248da9cc364b81ac67f5d67b626ee3f03a1f6299f7a9b4d0a0501548715365
 build:
-  script: export RUSTFLAGS="$RUSTFLAGS -C target-feature=+atomics,+bulk-memory,+mutable-globals"
+  script: |
+    rustup component add rust-src
```
ryanking13, line 13: @hoodmane WDYT about this? I am not sure if it is a good idea for individual package to run this command, as it will make global effect (AFAIK).

ryanking13: @hoodmane Do you think that `rustup component add rust-src` can overwrite the custom stdlib that we built with `wasm-exception` enabled?

Rust build toolchain is always confusing to me...

PR #320, opened 2025-08-18
bump inspice

pepijndevos:
Is there some automation to bump pypi packages?

This is only really here to test libngspice, and having it be outdated is unfortunate because it's a pure python package that could have just been installed from pypi


ryanking13:
> Well we used to have make -C packages update-all which was a bit more manual but worked alright.

Maybe we can have a similar command again, but as we now have more than 300 packages in this repository, I am not sure it would make sense to bump them all together.

PR #337, opened 2025-08-22
Add vrplib package

airen1986:


packages/vrplib/meta.yaml
```diff
@@ -0,0 +1,19 @@
+package:
+  name: vrplib
+  version: 2.0.0
+  top-level:
+    - vrplib
+source:
+  url: https://files.pythonhosted.org/packages/f4/a6/3cd94acf14036cee52418b4701caaf7bd0e0f619611648deb41725693e29/vrplib-2.0.0.tar.gz
```
ryanking13, line 7: vrplib is a pure python library. Could you use a wheel file directly instead of using the tar.gz archive?

ryanking13 (changes requested):

airen1986:
Hello @ryanking13 
I manually checked  sha-256 checksum for the vrplib/__init__.py file (at vrplib==2.0.0) is ac483bac89168e5ee4f13d10cef0afa38c68d8f8ae5c72478c9b44eb894ff2c5 , which is correctly stated in the RECORD file.Can you check the logic?
Also, created an issue with package maintainer: https://github.com/PyVRP/VRPLIB/issues/131 They also confirmed that the hash is correct.
Thanks

ryanking13:
Hmm, yeah I checked the sha256 hash for that file and the hash seems correct.

We do the hash check using the [pypa/wheel](https://github.com/pypa/wheel) tool. You can reproduce the error by

```
pip install wheel
python -m wheel unpack vrplib-2.0.0-py3-none-any.whl
# --> Unpacking to: .\vrplib-2.0.0...Hash mismatch for file 'vrplib/__init__.py'
```

ryanking13:
Oh, I found that the RECORD file of vrplib has sha256 hash in raw bytes, not in base64 encoded format. I am curious how it ended up having the raw hash. Note that [PEP-427](https://peps.python.org/pep-0427/#signed-wheel-files) required base64 format hash as far as I understand.

ryanking13:
Added a comment upstream, cross posting here too. I think it is an issue in uv-build side.

```
I found that the RECORD file of vrplib-2.0.0 has a raw SHA256 hash, but it should have a base64 encoded hash.

I think this is uv-build side issue. I found this upstream issue https://github.com/astral-sh/uv/issues/11723, which was fixed in February. Would you mind updating the uv-build version and make a new release?
```

ryanking13:
Yeah, it looks good. Thanks!

PR #339, opened 2025-08-23
Use dist_dir config variable instead of pyodide_root/dist

leehanjeong:
### Description
Related to https://github.com/pyodide/pyodide-build/issues/219

This updates scripts to use the new `dist_dir` config variable from pyodide-build, replacing the old pattern `$(pyodide config get pyodide_root)/dist`.

Now we can use: `$(pyodide config get dist_dir)`


leehanjeong:
Thanks for reviewing!
I updated the pyodide-build submodule as suggested, and the script using dist_dir var now works as expected!

ryanking13 (commented):
Thanks @leehanjeong!

ryanking13 (approved):

PR #341, opened 2025-08-25
add package: bilby.cython

ColmTalbot:
Migrating from https://github.com/pyodide/pyodide/pull/5502


packages/bilby.cython/meta.yaml
```diff
@@ -0,0 +1,22 @@
+package:
+  name: bilby.cython
+  version: 0.5.3
+  top-level:
+    - docs
```
ryanking13, line 5: Probably we can remove `docs` from top-level. I think it is too general import name for the package.

packages/bilby.cython/meta.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - bilby_cython
+source:
+  url: https://files.pythonhosted.org/packages/source/b/bilby_cython/bilby_cython-0.5.3.tar.gz
+  sha256: 44400b2abc6fa592b13d69f33460cb156d6edb48a783f5539ee0a21e1d0a3508
+requirements:
+  host:
+    - numpy
```
ryanking13, line 12: I guess numpy is not a build dependency. I cannot find it inside the pyproject.toml file of bilby_cython.

ColmTalbot: It's [here](https://git.ligo.org/colm.talbot/bilby-cython/-/blob/main/pyproject.toml?ref_type=heads#L7).

PR #342, opened 2025-08-26
Updated python-sat.

alexeyignatiev:
This PR updates the version of the *python-sat* package to `1.8.dev20`. The version (1) resolves the issue of duplicate clauses after clausification of non-clausal formulas is performed and (2) adds support for formulas compressed with zstd (to be brought by Python 3.14).


ryanking13 (approved):
Thanks. The test failure is not related to this PR.

PR #343, opened 2025-08-26
bilby.cython: remove bogus top level

ColmTalbot:
@ryanking13 thanks for catching this extra. I think the numpy build dependency is needed (it is for conda), but I'm happy to try removing it if you like?

### DIFF
--- a/packages/bilby.cython/meta.yaml
+++ b/packages/bilby.cython/meta.yaml
@@ -2,7 +2,6 @@ package:
   name: bilby.cython
   version: 0.5.3
   top-level:
-    - docs
     - bilby_cython
 source:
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Thanks, @ColmTalbot! Yes, noting that NumPy is required at build time here: https://git.ligo.org/colm.talbot/bilby-cython/-/blob/5a5f6028583f000252a3b086078d4145bb1fcbf4/pyproject.toml#L7

ryanking13 (commented):
Thanks @ColmTalbot!

Could you also update the test code for bilby.cython as well to use the exact package name? It was causing some issues in other PRs.

```diff
- @run_in_pyodide(packages=["bilby_cython"])
+ @run_in_pyodide(packages=["bilby.cython"])
```

PR #345, opened 2025-08-28
Bump python flint to 0.8.0

### LINKED ISSUE #322: Cython 3.1.0a1 is gone from PyPI and it is causing build issue in python-flint
As title.

I noticed that [python-flint](https://github.com/pyodide/pyodide-recipes/blob/main/packages/python-flint/meta.yaml) has a build issue recently, and the problem happens when installing the build requirement.

python-flint 0.7.1 (latest) has Cython 3.1.0a1 pinned in [pyproject.toml](https://github.com/flintlib/python-flint/blob/0.7.1/pyproject.toml), but for some reason, that version was deleted in PyPI. Hence, causing an error when installing the build dependency.

I think we need to make a patch to fix it.

- Related upstream commit: https://github.com/flintlib/python-flint/commit/6748eb2358929e4d25cb2d03ed7f2e3aa6fe8354

ryanking13:
Close #322

Applies upstream change: https://github.com/flintlib/python-flint/commit/6748eb2358929e4d25cb2d03ed7f2e3aa6fe8354

cc: @mkoeppe @oscarbenjamin


agriyakhetarpal:
Looks like the `bilby_cython` vs `bilby.cython` problem from #343 persists. :/

oscarbenjamin:
I've just pushed python-flint 0.8.0 to PyPI:
https://pypi.org/project/python-flint/

Maybe it makes more sense to just bump to that. It also supports FLINT 3.3 which I believe fixes some emscripten problems seen in https://github.com/flintlib/python-flint/pull/262

ryanking13:
@oscarbenjamin Sounds good. Thanks.

PR #350, opened 2025-09-02
Add python-calamine package

airen1986:


agriyakhetarpal (changes requested):
Thanks, @airen1986! It's building successfully. Could you please add a test?

PR #356, opened 2025-09-15
Make libproj shared and enable pyproj [full build]

tauroid:
See this discussion https://github.com/pyodide/pyodide/discussions/5887

TLDR: different instantiations of sqlite in different pyproj extension modules meant that sqlite was initialising global function pointers in one module and failing to find them in another. At least this is my take. Making libproj shared (and by extension, the statically linked sqlite inside libproj) appears to solve the issue.

~This PR is experimental for now until CI passes and I've thought about it a little more.~

Update: I bumped Cartopy and geopandas to check they pass too and looks good from here.

I can't see (as a new contributor) any more issues to solve. Apologies if I've missed something, let me know and I can change it.

### DIFF
--- a/packages/libgdal/meta.yaml
+++ b/packages/libgdal/meta.yaml
@@ -14,6 +14,7 @@ requirements:
     - libiconv
     - libgeos
   run:
+    - libproj
     - libgeos
 
 build:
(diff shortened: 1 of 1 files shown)


ryanking13 (commented):
Hmm, so do you think the problem is that sqlite3 is linked both in libproj and pyproj?

Then what happens if we just keep libproj static and remove sqlite3 from pyproj?

tauroid:
> Hmm, so do you think the problem is that sqlite3 is linked both in libproj and pyproj?

Not quite, I think it's that pyproj has 10 different extension modules (like `_context.cpython-313-wasm32-emscripten.so`, `_crs.cpython-313-wasm32-emscripten.so` etc), which each have a copy of libproj and sqlite3 statically linked inside, and it looks like libproj expects instead to be talking to one single "persistent" (in memory) sqlite library. In that it doesn't aggressively reinitialise sqlite whenever it calls it.

Though I'm not actually sure what the consequences of linking sqlite in libproj and then again in pyproj are 🤔 I guess the 10 extension modules could then also each have had some conflicting sqlite stuff inside? But I haven't personally seen any sign of that in what I've looked at so far.

tauroid:
> Then what happens if we just keep libproj static and remove sqlite3 from pyproj?

I can give this a quick try though, and revert if it fails (as I think it will)

tauroid:
Hmm okay, I did want to use rasterio, I'll look into it

PR #376, opened 2025-09-29
Bump imgui-bundle to v1.92.4

pthom:
This bumps imgui-bundle to v1.92.4

I am the author of ImGui Bundle. 

I just tested this package under pyodide and my test showed that this new version works.

Thanks!

### DIFF
--- a/packages/imgui-bundle/meta.yaml
+++ b/packages/imgui-bundle/meta.yaml
@@ -1,12 +1,12 @@
 package:
   name: imgui-bundle
-  version: 1.92.0
+  version: 1.92.4
   top-level:
     - imgui_bundle
 
 source:
-  url: https://github.com/pthom/imgui_bundle/releases/download/v1.92.0/srcs-full-v1.92.0.tar.gz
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks @pthom!

This PR looks good, but since you are a maintainer of this package, would you mind building the package in your local CI environment and providing a built wheel for this repository?

We are expecting that package maintainers will be able to upload Emscripten wheels to PyPI at some point ([PEP-783](https://peps.python.org/pep-0783/)). So we are asking maintainers to build and test WASM wheels in their repositories for future PyPI integration (and also to reduce the CI overhead in this repository).

ryanking13:
> However, I will need a way for me to be sure that pyodide users are able to install this package from a CDN, or from pypi.

Yes, you can still host it through Pyodide CDN or Pyodide package index until PyPI accepts WASM wheels. Reducing the CI runtime is enough of a goal for me.

After building the wheel in your CI, you can replace the URL in the recipe from the sdist file to the wheel that you built so that we don't need to rebuild it everytime.

PR #416, opened 2025-11-19
Enable XSPICE support in libngspice

### LINKED ISSUE #373: Compile ngspice with --enable-xspice
Hello! Thanks, @pepijndevos for adding ngspice with https://github.com/pyodide/pyodide-recipes/pull/102. (Here is the recipe: https://github.com/pyodide/pyodide-recipes/blob/main/packages/libngspice/meta.yaml.)

However, I'd like to compile with `--enable-xspice`, because I would like to use the Digital-to-Analog Node Bridge.

I'm happy to look into adding this functionality.

However, I'm having troubles building the existing package. Is there a dockerfile somewhere with the correct versions of pyodide, emsdk, python, etc? Or should I create it myself? Or is there a better way to build the package locally?

pepijndevos:
This fixes #373 supposedly, let's see what CI has to say about it. It's a bit sketchy with manually compiling native build dependencies.

Here is what Claude had to say about it:

> This commit enables XSPICE extensions in the libngspice recipe for WebAssembly, addressing several build challenges:
> 
> 1. Enable --enable-xspice configure flag
> 2. Build cmpp (code model preprocessor) natively first, then use it during the WebAssembly build to avoid cross-compilation issues
> 3. Add patch to fix verilog install-exec-hook which assumes shared libraries are built (not the case in WebAssembly builds)
> 
> The build process now:
> - Builds cmpp natively in build-native/ directory
> - Configures and builds WebAssembly version in release-lib/ directory
> - Replaces WebAssembly cmpp with native version before continuing build
> - Patches verilog Makefile to handle absence of shared libraries
> 
> Tested and verified to build successfully with XSPICE enabled.

### DIFF
--- a/packages/libngspice/meta.yaml
+++ b/packages/libngspice/meta.yaml
@@ -9,8 +9,8 @@ source:
   sha256: 3ae94703938247a22148f60a1307d52719d42a9134b5aef6252594000f513902
   patches:
     - patches/0001-keep-alive-API-functions.patch
-    - patches/0003-fix-verilog-install-hook.patch
     - patches/0002-fix-hicum2-extern-c.patch
+    - patches/0003-fix-verilog-install-hook.patch
 
 build:
   type: shared_library
@@ -51,7 +51,6 @@ build:
     cp ${WASM_LIBRARY_DIR}/lib/libngspice.so ${DISTDIR}
 
     # Copy XSPICE code model libraries to ngspice/*.cm
-    # Since shared_library packages extract to /usr/lib/, the final path will be /usr/lib/ngspice/*.cm
     mkdir -p ${DISTDIR}/ngspice
     cp ${WASM_LIBRARY_DIR}/lib/ngspice/*.cm ${DISTDIR}/ngspice/
 


pepijndevos:
Well after a lot of babysitting both Claude and my actual baby, I managed to coerce it towards the right solution. I wouldn't say it was efficient, but I could do it from my phone while also taking care of the actual baby.

The main things it struggled with are failing to build matplotlib to actually run the tests, and the CI error was from unpickeling exceptions that don't exist in the host, masking the actual error and sending it on a wild goose chase. Once I got it to test pyodide on node manually and add some error logging we started a very slow human+ci+ai in the loop debugging process.

packages/libngspice/meta.yaml
```diff
@@ -30,10 +37,24 @@ build:
...
+    cp ../build-native/src/xspice/cmpp/cmpp src/xspice/cmpp/cmpp
+    touch src/xspice/cmpp/cmpp
+
+    # Continue building the rest
     emmake make -j ${PYODIDE_JOBS:-3} LDFLAGS="${SIDE_MODULE_LDFLAGS}"
     emmake make install
     cp ${WASM_LIBRARY_DIR}/lib/libngspice.so ${DISTDIR}
 
+    # Copy XSPICE code model libraries to ngspice/*.cm
+    # Since shared_library packages extract to /usr/lib/, the final path will be /usr/lib/ngspice/*.cm
```
ryanking13:
```suggestion
```

This comment is not helpful as the installation path is useful only in runtime not in build time.

PR #417, opened 2025-11-25
matplotlib-inline 0.2.1, remove matplotlib dependency

bollwyvl:
Thanks for maintaining this!

Changes:
- [x] update to `matplotlib-inline 0.2.1`
- [x] remove the `run` dependency on `matplotlib` 
  - despite the name, `matplotlib-inline` [_doesn't_ depend on `matplotlib`](https://github.com/ipython/matplotlib-inline/blob/0.2.1/pyproject.toml#L34), and not every interactive session needs that big package (and `numpy`, and...) every time.

### DIFF
--- a/packages/matplotlib-inline/meta.yaml
+++ b/packages/matplotlib-inline/meta.yaml
@@ -1,12 +1,12 @@
 package:
   name: matplotlib-inline
-  version: 0.1.7
+  version: 0.2.1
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Thanks, @bollwyvl and @ianthomas23!

PR #419, opened 2025-11-28
Attempt to build tskit with numpy2

benjeffery:


benjeffery:
The artefacts in #418 appear to be built with numpy 1. I'm not sure how to make sure it is built with v2?

packages/tskit/meta.yaml
agriyakhetarpal: Hi @benjeffery, we already have NumPy v2 and all packages are using it for their builds. Could you please clarify?

If it helps, you can do something like this:


```suggestion
host:
  - numpy
constraint:
  - numpy > 2
```

PR #423, opened 2025-12-05
Add PyCDFpp package

jeandet:


packages/pycdfpp/meta.yaml
```diff
@@ -0,0 +1,19 @@
+package:
+  name: pycdfpp
+  version: 0.8.3
+  top-level:
+    - pycdfpp
+source:
+  url: https://github.com/SciQLop/CDFpp/archive/refs/tags/v0.8.3.tar.gz
```
ryanking13: We prefer downloading the wheel from PyPI rather than from GitHub commit. It seems like you don't upload Package sdist to PyPI, is there some reason behind it?

packages/pycdfpp/test_pycdfpp.py
```diff
@@ -0,0 +1,11 @@
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["pycdfpp", "requests"])
+def test_pycdfpp(selenium):
+    import pycdfpp
+    import requests
+    cdf = pycdfpp.load(requests.get('https://cdaweb.gsfc.nasa.gov/pub/software/cdawlib/0MASTERS/a1_k0_mpa_00000000_v01.cdf').content)
```
ryanking13: Could you change the unittest so that we don't make requests to an external website everytime? I think you can make a very small test file and load it from the filesystem instead.

ryanking13 (approved):
Thanks!

PR #430, opened 2025-12-12
Adds a few constraint for packages

ryanking13:
Fixes a few packages that were broken because of releases in their dependencies

cc @mkoeppe for primecountpy
cc @swnf for sxor

### DIFF
--- a/packages/primecountpy/meta.yaml
+++ b/packages/primecountpy/meta.yaml
@@ -18,6 +18,9 @@ requirements:
     - libprimesieve
   run:
     - cysignals
+  constraint:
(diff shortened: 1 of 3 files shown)


packages/primecountpy/meta.yaml
```diff
@@ -18,6 +18,9 @@ requirements:
     - libprimesieve
   run:
     - cysignals
+  constraint:
+    # ImportError: cysignals.signals does not export expected C function _do_raise_exception
+    - cysignals < 1.12.6
```
mkoeppe: This is https://github.com/passagemath/upstream-cysignals/issues/2 -- the people controlling cysignals broke the ABI in the middle of the release series

PR #438, opened 2025-12-16
Reenable pyarrow

ryanking13:
Re-enables pyarrow. I needed some hack to inject correct compiler while building... ideally we need to upstream this, but for now I am adding this script here so we can enable pyarrow faster (https://github.com/apache/arrow/pull/47330)

cc: @joemarshall @agriyakhetarpal


agriyakhetarpal:
Ah, makes sense. Thanks for clarifying! 🙏🏻

PR #439, opened 2025-12-19
Vendor shared libraries

ryanking13:
As proposed in https://github.com/pyodide/pyodide/issues/6030, I would like to enable shared library vendoring by default. This is a preparation for that to see if shared library vendoring works well for all existing packages.

### DIFF
--- a/packages/libopenblas/meta.yaml
+++ b/packages/libopenblas/meta.yaml
@@ -48,6 +48,9 @@ build:
 
     cp libopenblas.so dist
     emmake make install PREFIX=${WASM_LIBRARY_DIR}
+    # We need to copy the shared library again as the make install
+    # does not know we've modified the binary with libf2c symbols.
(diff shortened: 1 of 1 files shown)


packages/netcdf4/meta.yaml
```diff
@@ -13,19 +13,13 @@ requirements:
...
-    - libhdf5
 test:
   imports:
     - netCDF4
 build:
-  # FIXME(@ryanking13): In Pyodide 0.28, we have namespace collision issue when loading same shared library from multiple places.
-  # netCDF4 and h5py both depend on libhdf5, and if we vendor libhdf5 in both packages, the symbols will collide,
-  # and infinite recursion happens when loading the library.
-  # It is not clear how to fix this issue right now, so we disable vendoring for now.
-  # vendor-sharedlib: true
```
juntyr: I think I'm expericing this bug again, when libhdf5 is loaded twice. Do you remember what ended up fixing this bug?

PR #452, opened 2026-01-07
Add the google-crc32c package

juntyr:
This package will be required as a dependency for `zarr >= 3.1.4`

### DIFF
--- a/packages/google-crc32c/test_google_crc32c.py
+++ b/packages/google-crc32c/test_google_crc32c.py
@@ -1,6 +1,22 @@
 import pytest
 from pytest_pyodide import run_in_pyodide
 
+# adopted from https://github.com/googleapis/python-crc32c/blob/9652503fb296eef101d84b9b631f454cc7c1d538/tests/test___init__.py
+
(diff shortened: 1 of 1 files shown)


packages/libcrc32c/meta.yaml
```diff
@@ -0,0 +1,25 @@
+package:
+  name: libcrc32c
+  version: 1.1.0
```
ryanking13:
```suggestion
  version: 1.1.0
  tag:
    - library
    - shared_library
```

Let's add tags to help us manage recipes.

juntyr:
> Could you please add tests?

Done!

agriyakhetarpal (approved):
LGTM, thanks, @juntyr! Just one more change, please!

agriyakhetarpal (approved):
Awesome, thanks! I have a PR to make Zarr v3 work, but I haven't been actively tracking their dependencies often. I previously added the `crc32c` package back when Zarr needed it; looks like they've switched to `google-crc32c` now. Thanks again for keeping track of this.

PR #455, opened 2026-01-07
Add the numcodecs-wasm package

juntyr:
- `numcodecs-wasm` is the root package for my https://numcodecs-wasm.readthedocs.io/en/latest/ project, which provides several compression algorithms compiled to WASM. `numcodecs-wasm` provides the extension module to run the compressors, while all other packages, e.g. `numcodecs-wasm-sz3`, are pure-Python packages that can be installed on-demand with `micropip`
- `numcodecs-combinators` is a helper package for writing meta-compressors with `numcodecs`
- `numcodecs-observers` is a meta-compressor for recording performance stats during compression, `numcodecs-wasm` hooks into it to provide an additional performance stat (an instruction count measured inside WASM)
- `varint` is required by `numcodecs-combinators`


hoodmane:
Could you add tests?

packages/varint/meta.yaml
```diff
@@ -0,0 +1,13 @@
+package:
+  name: varint
+  version: 1.0.2
+  top-level:
+    - varint
+source:
+  url: https://files.pythonhosted.org/packages/source/v/varint/varint-1.0.2.tar.gz
```
ryanking13: Do you really need this package as a dependency?

This package is just a 65 lines of code (https://github.com/fmoo/python-varint/blob/master/varint.py) that was updated 8 years ago. Since you are the owner of the packages in this PR, I think you should be able to remove this package.

packages/numcodecs-wasm/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+  script: |
+    # FIXME: remove once https://github.com/pyodide/pyodide/pull/6027 is used in pyodide-recipes
+    export RUSTUP_TOOLCHAIN="nightly-2025-12-10"
+    rustup toolchain install ${RUSTUP_TOOLCHAIN}
+    rustup target add wasm32-unknown-emscripten --toolchain ${RUSTUP_TOOLCHAIN}
+about:
+  home: https://github.com/juntyr/numcodecs-rs
+  PyPI: https://pypi.org/project/numcodecs-wasm
+  summary: numcodecs compression for codecs compiled to WebAssembly
+  license: MPL-2.0
```
ryanking13: Could you add yourself as a package maintainer (same for other recipes as well)?

packages/varint/meta.yaml
juntyr: I'm not entirely sure about that. I use this package in quite a few dependencies, so I would not want to vendor it separately for each, so at best I could copy-paste the code, republish it to PyPi with a pure Python wheel, and control the dependency. But that also feels a bit silly. It's a small functionality, the implementation is quite decent (and most importantly minimal), so I've actually come to like the varint package :D

juntyr:
> Could you add tests?

Absolutely! The easiest way to test would likely be to use micropip to install (so importantly we wouldn't need to store the WASM file in this repo since it's hosted on PyPi) one of the actual codecs from PyPi and let that encode + decode some data, which exercises all code in numcodecs-wasm. Does that sound good?

agriyakhetarpal:
With PEP 783 accepted, could a path forward be to close this PR and attempt building in-tree, such that you can reuse the existing test suite – which was what was remaining here?

juntyr:
I already build and test the package for Pyodide in my CI, I just haven't issued a new release yet

juntyr:
numcodecs-wasm v0.2.5 is now on PyPi with Pyemscripten 2026_0 wheels: https://pypi.org/project/numcodecs-wasm/0.2.5/#files

PR #456, opened 2026-01-07
Add recipe for Pyrodigal

althonos:
This PR adds a recipe for Pyrodigal.


hoodmane:
Could you add a test?

althonos:
Is there a way to specify a way to run the unit tests? The package is distributed with its own tests so it should be possible to run just `python -m unittest pyrodigal.tests`.

althonos:
Thank you!

PR #482, opened 2026-01-21
Upgrade numcodecs to v0.15.1

juntyr:
numcodecs v0.15.1 is not the latest version (that's v0.16.x), but it's the latest one that is compatible with Zarr v2.x, which is still widely used, while also being compatible with Zarr v3.x

### DIFF
--- a/packages/numcodecs/test_numcodecs.py
+++ b/packages/numcodecs/test_numcodecs.py
@@ -2,7 +2,7 @@ from pytest_pyodide import run_in_pyodide
 
 
 @run_in_pyodide(
-    packages=["lzma", "numcodecs", "numpy"],
+    packages=["numcodecs", "numpy"],
 )
 def test_blosc(selenium_standalone):
     import array


packages/numcodecs/test_numcodecs.py
```diff
@@ -1,12 +1,8 @@
-import pytest
 from pytest_pyodide import run_in_pyodide
 
 
-@pytest.mark.xfail_browsers(
-    chrome="test_numcodecs triggers a recursion error in chrome"
-)
 @run_in_pyodide(
-    packages=["numcodecs", "numpy"],
```
juntyr: numcodecs expects that the lzma builtin module exists but Pyodide unvendors it. Should/can numcodecs depend on lzma as a run requirement, to ensure it is loaded if the users did not preload it, or is adding it as a test dependency here sufficient?

juntyr: just adding it to the packages here clearly doesn't work ... @ryanking13 how do I preload lzma for numcodecs (tests)?

PR #485, opened 2026-01-21
Add coolname package

jolaf:


packages/coolname/meta.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - always
+  top-level:
+    - coolname
+
+source:
+  sha256: 4d1563186cfaf71b394d5df4c744f8c41303b6846413645e31d31915cdeb13e8
+  url: https://files.pythonhosted.org/packages/1b/b1/5745d7523d8ce53b87779f46ef6cf5c5c342997939c2fe967e607b944e43/coolname-2.2.0-py2.py3-none-any.whl
+
+about:
+  home: https://github.com/alexanderlukanin13/coolname
```
agriyakhetarpal: It looks like the package hasn't been touched for a few years. The maintainer has not been active on GitHub either. While it will still continue to work as it is not a complex package, I wanted to ask two things:
- Could you please take up maintainership of this recipe? You may refer to other recipes to see how to add yourself as a recipe maintainer here. The idea is that we will know whom to contact if there are any test failures or other issues, and you could help us with those.
- If we require any patches to fix any code or new behaviour, then it is likely that you might need to fork the package under a new name, say `coolname-ng` or `coolname2`, and add its recipe here instead. Would you be willing to help with that?

packages/coolname/meta.yaml
```diff
@@ -0,0 +1,16 @@
...
+  PyPI: https://pypi.org/project/coolname
+  license: BSD
```
agriyakhetarpal: Let's use the SPDX identifier for the license instead: https://spdx.org/licenses/BSD-2-Clause.html
```suggestion
  license: BSD-2-Clause
```

agriyakhetarpal (changes requested):
Hi @jolaf, thanks for adding a recipe for `coolname`! This is a good start. Could you please add some tests? You may look at how other packages have their tests set up next to their recipe files, but the idea is that we want to have a few basic checks that the package is importable, and rudimentary parts of its functionality work as expected and run in our CI.

PR #493, opened 2026-02-02
Phispy

linsalrob:
[PhiSpy](https://github.com/linsalrob/PhiSpy) is a prophage finder that identifies prophages in bacterial and archaeal genomes using multiple metrics.

Package includes:
 - C++ extension module (PhiSpyRepeatFinder) for efficient repeat finding
 - Dependencies: biopython, numpy, scikit-learn (all available in Pyodide)

Tested with PhiSpy 5.0.1 from PyPI.

### DIFF
--- a/packages/phispy/test_phispy.py
+++ b/packages/phispy/test_phispy.py
@@ -5,7 +5,7 @@ def test_phispy_version(selenium):
     import PhiSpyModules
     
     version = PhiSpyModules.__version__
-    assert version == "5.0.1"
+    assert version == "5.0.2"
 
 
 @run_in_pyodide(packages=["phispy"])


linsalrob:
PhiSpy v. 5.0.2 introduces lazy loading for the module that fails

packages/phispy/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  url: https://files.pythonhosted.org/packages/d4/6b/cbc4b6be7755c7a1ffd58e51e878a57713f00829d340a46e894061477fc0/phispy-5.0.2.tar.gz
+  sha256: a9233471ccf50e6e58ecc83c589be2cfa04e728cc81e57cea296d23dc85d397c
+
+build:
+  cxxflags: -fwasm-exceptions
+  ldflags: -fwasm-exceptions
+
+requirements:
+  host:
+    - numpy
```
ryanking13: Do you use numpy while building the package? Looking at the [`pyproject.toml`](https://github.com/linsalrob/PhiSpy/blob/2436a052951abb3f319990756c2e1a76d894595d/pyproject.toml#L1-L3) file, it doesn't seem to use numpy in the build process. If you are not using numpy during the build steps, you can remove this line.

ryanking13 (commented):
Could you please add a basic unittest for the package? Testing a very basic scenario would work such as https://github.com/pyodide/pyodide-recipes/blob/main/packages/jedi/test_jedi.py

PR #504, opened 2026-02-16
.github/workflows/build.yml: Upload artifacts also on failing builds

mkoeppe:
... for inspection when builds fail

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -250,6 +250,7 @@ jobs:
 
       - name: Parse build results
         id: parse_results
+        if: always()
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
ryanking13: This script does not handle build failure. If it is for debugging, running this script won't be very useful.

ryanking13 (changes requested):
Sound reasonable, but the build logs are not copied to `build-logs` directory unless the build succeeds, so this PR will not work. I guess we need to the behavior of `pyodide-build` first to make this work.

PR #554, opened 2026-04-03
Fix healpy recipe

juntyr:
In Emscripten 4 `-sRELOCATABLE=1` was required in `healpix_cxx` to link with PIC-enabled libc. In Emscripten 5, `-sSIDE_MODULE=1` must be set instead.


juntyr:
I decided to try to fix this on my work, which still runs on Pyodide v0.29 (so Emscripten v4.0.9) but also had the same issue when relocatable was turned off. Here's the full build log with verbose make commands:

[build-2.txt](https://github.com/user-attachments/files/26596633/build-2.txt)

The link command that fails ends up being the following:

```
libtool: link: /src/pyodide-recipes/packages/healpy/build/pywasmcross_symlinks/c++  -fPIC -DPIC -shared -nostdlib   cxxsupport/.libs/announce.o cxxsupport/.libs/geom_utils.o cxxsupport/.libs/string_utils.o cxxsupport/.libs/ls_image.o cxxsupport/.libs/paramfile.o cxxsupport/.libs/pointing.o cxxsupport/.libs/rotmatrix.o cxxsupport/.libs/trafos.o cxxsupport/.libs/walltimer.o cxxsupport/.libs/wigner.o cxxsupport/.libs/error_handling.o cxxsupport/.libs/fitshandle.o Healpix_cxx/.libs/alm.o Healpix_cxx/.libs/alm_healpix_tools.o Healpix_cxx/.libs/alm_powspec_tools.o Healpix_cxx/.libs/healpix_tables.o Healpix_cxx/.libs/healpix_base.o Healpix_cxx/.libs/healpix_map.o Healpix_cxx/.libs/powspec.o Healpix_cxx/.libs/moc_query.o Healpix_cxx/.libs/weight_utils.o Healpix_cxx/.libs/mask_tools.o Healpix_cxx/.libs/alice3.o Healpix_cxx/.libs/healpix_data_io.o Healpix_cxx/.libs/healpix_map_fitsio.o Healpix_cxx/.libs/alm_fitsio.o Healpix_cxx/.libs/powspec_fitsio.o Healpix_cxx/.libs/moc_fitsio.o   -Wl,-rpath -Wl,/src/pyodide-recipes/packages/healpy/build/healpy-1.19.0/build/temp.emscripten_4_0_9_wasm32-cpython-313/lib -Wl,-rpath -Wl,/src/pyodide-recipes/packages/healpy/build/healpy-1.19.0/build/temp.emscripten_4_0_9_wasm32-cpython-313/lib -L/src/cpython/installs/python-3.13.2/lib/ -L/src/pyodide-recipes/packages/healpy/build/healpy-1.19.0/build/temp.emscripten_4_0_9_wasm32-cpython-313/lib /src/pyodide-recipes/packages/healpy/build/healpy-1.19.0/build/temp.emscripten_4_0_9_wasm32-cpython-313/lib/libsharp.so -lm /src/pyodide-recipes/packages/healpy/build/healpy-1.19.0/build/temp.emscripten_4_0_9_wasm32-cpython-313/lib/libcfitsio.so -L/src/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten -L/src/emsdk/emsdk/upstream/emscripten/src/lib -lGL-getprocaddr -lal -lhtml5 -lstubs -lnoexit -lc -ldlmalloc -lcompiler_rt-legacysjlj -lc++-legacyexcept -lc++abi-legacyexcept -lunwind-legacyexcept -lsockets   -g -O3 -O2 -g0 -O2 -g0   -Wl,-soname -Wl,libhealpix_cxx.so.4 -o .libs/libhealpix_cxx.so.4.0.5
```

ryanking13:
Try adding `-sSIDE_MODULE=1` in cxxflags or ldflags or both (in Emscripten 5.0.3). It should already be added, but just in case.

The problematic code line is here: https://github.com/emscripten-core/emscripten/blob/bbf1caa6e24f64fca9eb6a13a9e02d3f42123e77/tools/cache.py#L125-L126, this branch should hit and the directory that searches the library should include the `/pic/`.

In Emscripten 4.0.9, it checked `RELOCATABLE` flag. In Emscripten 5.0.X, it was changed to check the SIDE_MODULE flag as RELOCATABLE flag was removed. Probably SIDE_MODULE is not set in your link command for any reason (I doubt it but still)

juntyr:
> Try adding `-sSIDE_MODULE=1` in cxxflags or ldflags or both (in Emscripten 5.0.3). It should already be added, but just in case.

Thanks, that seems to have worked, and the final diff is nice and clean (especially when compared to the version before you disabled the recipe).

juntyr:
I was happy too soon ...

ryanking13:
Well at least the build passed now, so it is a progress 😅.

> TypeError: Cannot read properties of undefined (reading 'value')

This is a regression that I experienced in [libmagic](https://github.com/pyodide/pyodide-recipes/blob/8c76670ae1981a003c4bf06f0b8df986467be50d/packages/libmagic/meta.yaml#L4) as well after upgrading the Emscripten version. I haven't been able to spot the root cause yet though. If you could be able to debug the code pointer that the error is happening, it would be wonderful. I believe that the error is happening somewhere inside [libdylink.js](https://github.com/emscripten-core/emscripten/blob/main/src/lib/libdylink.js).

ryanking13:
I guess this was related to Pyodide-side regression (https://github.com/pyodide/pyodide/pull/6200)

juntyr:
Thank you for investigating! Let's see what CI says once your fix has landed

agriyakhetarpal:
Since https://github.com/pyodide/pyodide/pull/6200 landed in 314.0.0a2, I wonder if the failure here is just because we are using an older xbuildenv here, which corresponds to 314.0.0a1:

https://github.com/pyodide/pyodide-recipes/blob/e70bb6198b03cc7c569b11f3a9992f3e9b21f37b/pyproject.toml#L84

And if we drop it (which will then use 314.0.0 stable), the build will be fixed? I walked through the commits here one-by-one and I noticed that this hasn't been tried out, so I think it is worth trying.

agriyakhetarpal:
Oh hey, all green and the test is passing! @juntyr, please feel free to take this up to the healpy developers too – maybe they could be open to publishing pyemscripten wheels per PEP 783 now that 314.0.0 is out? Otherwise I am fine with this being here too, obviously, since it's a pre-existing recipe.

juntyr:
@agriyakhetarpal Thank you for making it work! I'm quite swamped at the moment, so I think merging this would be easier. Since I'm also maintaining this recipe in my fork, I'll be motivated later to try to upstream this :)

agriyakhetarpal (approved):
No problem, thanks for stopping by here to attempt this fix! Thanks to you too @ryanking13!

PR #557, opened 2026-04-04
Add segno package recipe and runtime tests

anxkhn:
## Summary

- add a new `segno` recipe (`1.6.6`) as a pure Python wheel
- set package metadata (`home`, `PyPI`, `summary`, `license`, maintainer)
- add import smoke test in recipe metadata
- add package runtime tests in `packages/segno/test_segno.py`

## Why

`segno` is a useful QR Code and Micro QR Code generator that fits Pyodide packaging well:

- pure Python distribution
- no native extension dependency chain
- useful for browser-only workflows and demos

## Testing

Built recipe locally:

- `pyodide build-recipes segno --install --log-dir build-logs`

Prepared runtime:

- `./tools/copy_pyodide_runtime.sh ./dist`

Ran package tests:

- `pytest -v --dist-dir=./dist --runner=selenium --rt chrome packages/segno/test_segno.py`

Result:

- `2 passed`

Closes #556

### DIFF
--- a/packages/segno/meta.yaml
+++ b/packages/segno/meta.yaml
@@ -1,0 +1,19 @@
+package:
+  name: segno
+  version: 1.6.6
+  top-level:
+    - segno
+source:
+  url: https://files.pythonhosted.org/packages/py3/s/segno/segno-1.6.6-py3-none-any.whl
+  sha256: 28c7d081ed0cf935e0411293a465efd4d500704072cdb039778a2ab8736190c7
+about:
+  home: https://github.com/heuer/segno/
+  PyPI: https://pypi.org/project/segno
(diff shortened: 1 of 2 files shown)


ryanking13:
As this is a pure python package, you should be able to install it directly from PyPI without adding it to Pyodide lockfile. Is there a specific reason you want to add it here?

agriyakhetarpal:
Let's close this PR, as we have not received any response from @anxkhn for some time. @anxkhn, if you think `segno` needs patches to work in Pyodide or attention in Pyodide's packaging context, please let us know and we can explore reopening this! For now, I'll assume that `segno` works fine in Pyodide, since it can be installed with micropip as a pure Python wheel. Thank you!

PR #559, opened 2026-04-07
Add jsonschema-rs package

evroon:
Adds the [jsonschema-rs](https://github.com/Stranger6667/jsonschema/tree/master/crates/jsonschema-py) package

I am aware there is already a jsonschema package in pyodide but it doesn't have proper typing (and is less performant than jsonschema-rs).

See discussion regarding how to build this: https://github.com/pyodide/pyodide/discussions/6178

### DIFF
--- a/packages/jsonschema-rs/meta.yaml
+++ b/packages/jsonschema-rs/meta.yaml
@@ -19,6 +19,9 @@ build:
 requirements:
   executable:
     - rustup
+test:
+  imports:
+    - jsonschema_rs
 about:
   home: https://github.com/Stranger6667/jsonschema
   PyPI: https://pypi.org/project/jsonschema-rs/0.45.0
--- a/packages/jsonschema-rs/meta.yaml
+++ b/packages/jsonschema-rs/meta.yaml
@@ -1,0 +1,29 @@
+package:
+  name: jsonschema-rs
+  version: 0.45.0
+  tag:
+    - rust
+  top-level:
+    - jsonschema_rs
+source:
+  url: https://files.pythonhosted.org/packages/d2/7b/40ed0aa40ff2f3bc9bfccf20ea29d9a99199a8eefda26ae9d65601c144fa/jsonschema_rs-0.45.0.tar.gz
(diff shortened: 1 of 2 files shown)


packages/jsonschema-rs/meta.yaml
```diff
@@ -0,0 +1,32 @@
+package:
+  name: jsonschema-rs
+  version: 0.45.0
+  tag:
+    - rust
+  top-level:
+    - jsonschema_rs
+source:
+  url: https://files.pythonhosted.org/packages/d2/7b/40ed0aa40ff2f3bc9bfccf20ea29d9a99199a8eefda26ae9d65601c144fa/jsonschema_rs-0.45.0.tar.gz
```
agriyakhetarpal: PyPI has predictable URLs for sdists (and wheels, too), so this can be simplified as follows:

```suggestion
  url: https://files.pythonhosted.org/packages/source/j/jsonschema_rs/jsonschema_rs-0.45.0.tar.gz
```

If you use `pyodide skeleton` to create the recipe, it will automatically resolve the URLs. :D

agriyakhetarpal (commented):
Thanks for contributing, @evroon! This is an interesting package; I didn't know something like it existed. I'm going to find it pretty useful for my side projects and other repos I contribute to. :)

Could you please add a test or two? They don't have to be complicated; just a few tests from the repository's test suite that ensure basic functionality works should suffice.

ryanking13:
Thanks for the contribution @evroon. Actually, would you like to contact to the jsonschema-rs maintainer and ask to include pyodide build in their ci and publish it to PyPI directly? We've been maintaining this repository as there was no way to host WASM wheels before. But we just got [PEP 783](https://peps.python.org/pep-0783/) accepted (yesterday!) and it will soon (in a few days) possible to publish wasm wheels to Pyodide. So we do not want to maintain these packages ourselves anymore.

It should be good for you as well as if you add the package here, we will ping you anytime we have issues with this pacakge.

evroon:
Congrats on the PEP! Yes I'll open an issue in that repo and potentially make a PR. I guess I should take a project like `polars` as an example to follow on setting this up?

PR #562, opened 2026-04-16
Add `uuid-utils`

aminalaee:


aminalaee:
I'll close this and open a new one for version 0.15.0 which will support pyodide

ryanking13:
@aminalaee Thanks! I would say we are not actively accepting new recipes in this repository anymore as [PEP 783](https://peps.python.org/pep-0783/) is accepted and you are now possible to deploy your wheels directly to PyPI. it means you can install your package without adding it here.

(To be more precise, we are waiting for PyPI to accept pyemscripten wheels (https://github.com/pypi/warehouse/pull/19804), but it should be done in a few days)

PR #575, opened 2026-05-03
Bump cryptography v47

greateggsgreg:
Also working on getting upstreamed support for builds here: https://github.com/pyca/cryptography/issues/14788


ryanking13:
Awsome. Thanks for upstreaming it.

greateggsgreg:
To get this to work, we'll need to build and link OpenSSL 3.0+ with this wheel. I see that openssl is being dropped from the default build in 314.0. Is the intent that all wheels that need OpenSSL will build it themselves? Is the intent to drop packages from pyodide-recipies if we can get things upstreamed? Might need to fix up the build here if cryptography doesn't merge our PR in time for the 314.0 release but welcome recommendations on where effort is best placed.

Started other upstreaming efforts here:
https://github.com/pyca/pynacl/issues/946
https://github.com/pydantic/pydantic/pull/13157

ryanking13:
> I see that openssl is being dropped from the default build in 314.0. Is the intent that all wheels that need OpenSSL will build it themselves?

Yes.

> Is the intent to drop packages from pyodide-recipies if we can get things upstreamed?

Partially, yes. We can still have the built wheel in the pyodide-recipes to make new lockfiles compatible to the old one, but my plan is slowly dropping packages and make the lockfile optional in the future.

>  Might need to fix up the build here if cryptography doesn't merge our PR in time for the 314.0 release but welcome recommendations on where effort is best placed.

Sure, PR welcome for patching existing packages to make it work.

ryanking13 (approved):
Thanks! Let's consider cleaning up the recipe (or removing it from here) when your upstream PR is merged.

PR #578, opened 2026-05-11
Matplotlib: separate pyodide backend instead of patched webagg

### LINKED ISSUE #475: Matplotlib patched webagg backend breaks ipympl
I patched the `matplotlib` `pyodide` recipe in pyodide/pyodide#5398 so that `webagg` is an interactive backend available in standalone `pyodide` deployments, i.e. outside of JupyterLite and not in a WebWorker.

Unfortunately it breaks `ipympl` which is the interactive `matplotlib` backend used in the pyodide kernel in JupyterLite. With the benefit of hindsight this is obvious as `ipympl` inherits from some of the patched code, but I did not consider it at the time.

Options are:
1. Do nothing. This is great for standalone pyodide but bad for the pyodide JupyterLite kernel.
2. Remove the patch completely. This is great for the pyodide JupyterLite kernel but bad for standalone pyodide.
3. Rearrange the patch to try to work for both. I don't know how much work this is.

Option 3 is the only sensible option here, but I am aware that at some date in the future the patches in `pyodide-recipes` become obsolete as individual projects like Matplotlib will be building and publishing their own wheels, so the projects themselves will have to carry patches for the pyodide build. Is there a concrete timeline for this? It feels to me like that is more than a couple of months away?

Also pinging @tacaswell about how this backend might end up within the `matplotlib` codebase, probably as a separate backend (which you did suggest at the time)? It will mean adding more JavaScript or even TypeScript to the codebase.

ianthomas23:
## Description

This PR fixes the use of Matplotlib `ipympl` when using the pyodide kernel in JupyterLite. Fixes #475.

## Type of change

- [ ] Adding a new package
- [x] Bug fix
- [ ] Other (please describe)

## Notes

Previously in pyodide/pyodide#5398 I patched the Matplotlib `webagg` backend to give an interactive backend that works in pyodide when running outside of JupyterLite, but within JupyterLite the patched code breaks other web-based backends such as `ipympl`.

Here instead of patching the `webagg` backend I move the code to a new `pyodide` backend to leave the base `webagg` backend code unchanged so that other uses of it continue to work as expected. 

The new patch file `0003-Pyodide-webagg-backend.patch` can be compared to the unpatched Matplotlib code at https://github.com/matplotlib/matplotlib/compare/v3.10.8...ianthomas23:matplotlib:v3.10.8-pyodide-backend?expand=1. `backend_webagg.py` is now unchanged and there are only some small simple changes in `backend_webagg_core.py` and `mpl.js`. New files `backend_pyodide.py` and `mpl_pyodide.js` override the base class behaviour, the former in the usual manner and the latter by being loaded after the base `mpl.js` to overwrite previously loaded functions. The backend is now called the `pyodide` backend but I have defined `webagg` as a synonym for this and kept `webagg` as the default when running in pyodide so that if you do not specify a backend or you use `import matplotlib as mpl; mpl.use('webagg')` you get exactly the same behaviour (outside of JupyterLite) as before this PR.

Example screenshot below when using the default backend in this code:
```html
<!doctype html>
<html>
  <head>
    <script src="pyodide.js"></script>
  </head>
  <body>
    <script type="text/javascript">
      async function main(){
        let pyodide = await loadPyodide();
        await pyodide.loadPackage("micropip");
        const micropip = pyodide.pyimport("micropip");
        await micropip.install('matplotlib');
        await pyodide.runPython(`
          import matplotlib as mpl
          import matplotlib.pyplot as plt
          import numpy as np
          fig, ax = plt.subplots()
          x, y = np.meshgrid(np.linspace(0, 2*np.pi, 20), np.linspace(0, 2*np.pi, 20))
          z = np.sin(x - y/4)*np.sin(y)
          cs = ax.contourf(x, y, z)
          fig.colorbar(cs, ax=ax)
          fig.suptitle(f"backend: {mpl.get_backend()}")
          plt.show()`
        );
      }
      main();
    </script>
  </body>
</html>
```

Screenshot 2026-05-11 at 09 59 11

Given that Matplotlib is now building its own pyodide wheels in CI, this code or something like it needs to be in the Matplotlib source so after this is merged I will look at doing that. No doubt the outcode won't be exactly the same as here as there will need to be discussion and agreement on the code and the naming, but it should be very similar functionally.


packages/matplotlib/meta.yaml
```diff
@@ -22,19 +22,14 @@ build:
     setup-args=-DrcParams-backend=webagg
     setup-args=-Dsystem-freetype=true
   cflags: |
-    -s USE_FREETYPE=1
```
ianthomas23, line 25: Various `USE_*` flags here are no longer needed now that Matplotlib is built using `meson`.

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
```diff
@@ -0,0 +1,353 @@
...
+@@ -0,0 +1,152 @@
++"""Interactive backend for pyodide running in main browser thread, based on webagg."""
++
++import base64
++from io import BytesIO
++import json
++import mimetypes
++from pathlib import Path
++
++from js import alert, document
```
ryanking13: Could you please move the `document` import into the place where it is actually used and guard it with `try ... except`?

In Node.js, there is no document object, and there were some packages that depends on matplotlib having ImportError when run in Node.js.

Users will get runtime error when they try to access the document anyways, but at least we can prevent from having ImportError at the initialization stage. Also the codepaths that does not use the `document` object directly will still work in Node.js

ryanking13 (approved):
> Given that Matplotlib is now building its own pyodide wheels in CI, this code or something like it needs to be in the Matplotlib source so after this is merged I will look at doing that. No doubt the outcode won't be exactly the same as here as there will need to be discussion and agreement on the code and the naming, but it should be very similar functionally.

Awesome! Thanks for doing that @ianthomas23. Feel free to ping me if you need any assist from Pyodide side.

Codewise looks good to me. I added one comment about Node.js issue.

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
agriyakhetarpal: Probably `alert` too? I saw a use of it in `handle_save`.

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
```diff
@@ -0,0 +1,353 @@
...
++
++        def send_json(self, content):
++            self.js_web_socket.receive_json(json.dumps(content))
++
++        def send_binary(self, blob):
++            if self.supports_binary:
++                self.js_web_socket.receive_binary(blob, binary=True)
++            else:
++                data_uri = "data:image/png;base64,{}".format(
++                    blob.encode('base64').replace('\n', ''))
```
agriyakhetarpal, line 149: I guess we should do `base64.b64encode(blob).decode('ascii')`? This is Python-2-only, if I remember correctly.

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
```diff
@@ -0,0 +1,353 @@
...
++        const image_bytes = mpl.toolbar_image_callback(image_name).toJs({create_pyproxies : false});
++        const blob = new Blob([image_bytes], { type: 'image/png' });
++        blob_url = URL.createObjectURL(blob);
++        this._toolbar_urls[image_name] = blob_url;
++    }
++
++    const icon_img = new Image();
++    icon_img.src = blob_url;
++    return icon_img;
++};
```
agriyakhetarpal, line 360: I _think_ the cache is a no-op. Because, as written, `this._toolbar_urls` is per-figure, but the toolbar images are all distinct [image_names](https://github.com/matplotlib/matplotlib/blob/df808aae3e1b884c35ebfbd64ef58e7420958ff4/lib/matplotlib/backends/backend_webagg_core.py#L405-L413), so the cache never hits within a figure. A fresh figure starts an empty cache, so it doesn't get shared across figures either. Since your intent based on the comment is page-wide sharing, we should move it to `mpl._toolbar_urls`, otherwise it can be removed?

agriyakhetarpal (approved):
Great, thanks @ianthomas23 for your work!

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
ianthomas23: Yes, you are correct that the cache is a no-op. I made a few improvements in passing and I have optimised it to do nothing :smile:

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
ianthomas23: I don't understand your reference to Python-2.

This is a copy of existing code already in Matplotlib in two places, and I particularly want to reuse the existing code as much as possible.

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
ianthomas23: I've done this in 83061e8, or you can see it as a separate commit at https://github.com/ianthomas23/matplotlib/commits/v3.10.8-pyodide-backend/

For the `handle_save` I've converted the `ImportError` to a `RuntimeError` to add some more context. But it could just remain as an `ImportError` if you'd prefer?

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
ianthomas23: I've switched it to `mpl._toolbar_urls` as you suggested, in 83061e8 or as a separate commit at https://github.com/ianthomas23/matplotlib/commits/v3.10.8-pyodide-backend/

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
agriyakhetarpal: What I mean is this:

```python
b"test".encode("base64")
```

```
'bytes' object has no attribute 'encode'
```

That works only in Python 2. In Python 3, this should be:

```python
import base64
base64.b64encode(b"test").decode('ascii')
```

```python
'dGVzdA=='
```


---

After you suggested that this is part of existing code in Matplotlib, I checked:

`backend_nbagg.py`, i.e., https://github.com/matplotlib/matplotlib/blob/1392cbe3c79cdb93f9282747841d648770f60249/lib/matplotlib/backends/backend_nbagg.py#L213-L221 has the correct, Python 3 form

`backend_webagg.py`, i.e, https://github.com/matplotlib/matplotlib/blob/1392cbe3c79cdb93f9282747841d648770f60249/lib/matplotlib/backends/backend_webagg.py#L159-L164 does not and still uses the Python 2 form. I assume no one hit this path in practice, really, so no one ever fixed it, and it ended up surviving in this form 😄

It is still the same on `main` right now: https://github.com/matplotlib/matplotlib/blob/97fb4d494a159ce4ea6298cff3f02fcddee97885/lib/matplotlib/backends/backend_webagg.py#L159-L164, so it's something to fix upstream, I guess! No action needed here though!

---

P.S. I'm happy to do it if you'd like me to; I'm always a fan of hunting down things like this and I don't think I've contributed to Matplotlib before, so it'd be nice to get my first PR in. And I can run whatever moderniser tool I can find to catch more such instances, of course.

Edit: I couldn't recall the name immediately, but [pyupgrade](https://github.com/asottile/pyupgrade) will do it. Looks like it was run [once](https://github.com/search?q=repo:matplotlib/matplotlib+%22pyupgrade%22&type=code), in https://github.com/matplotlib/matplotlib/commit/ffbb4c151cd0e102d47bf74189e39480f5f8f30e, and it didn't catch this case at the time for whatever reason.

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
agriyakhetarpal: I think `RuntimeError` makes sense!

packages/matplotlib/patches/0003-Pyodide-webagg-backend.patch
ianthomas23: OK, I see now. By default the various webagg-related backends have `supports_binary = True` so this code path in both Matplotlib and this patch are never used, and this code could be replaced with anything. Maybe it would be better to replace it with throwing a "Not implemented" runtime exception rather than what looks like useful code.

On the other hand the nbagg backend has `supports_binary = None` so this code path is used and that is presumably why it was fixed in the past.

Go ahead and change it in Matplotlib if you wish. Note that there is very little coverage of the webagg/nbagg backends there (https://app.codecov.io/gh/matplotlib/matplotlib/tree/main/lib%2Fmatplotlib%2Fbackends) and that is unlikely to change.

agriyakhetarpal: Interesting, thanks for all the context! I'll prepare a PR when I can get some time to! I guess we're one of the more prominent users of the WebAgg backend nowadays, given that matplotlib-pyodide isn't used anymore.

PR #579, opened 2026-05-13
jsonschema basic tests

greateggsgreg:
Adds basic pytest coverage for `jsonschema` (import, validate, format-checker, Draft202012Validator).

### DIFF
--- a/packages/jsonschema/test_jsonschema.py
+++ b/packages/jsonschema/test_jsonschema.py
@@ -1,0 +1,56 @@
+import pytest
+from pytest_pyodide import run_in_pyodide
+
+
+@run_in_pyodide(packages=["jsonschema"])
+def test_jsonschema_import_draft202012(selenium):
+    from jsonschema import Draft202012Validator
+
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (commented):
Thanks @greateggsgreg! If you think this is useful to include otherwise, I am also fine with adding an xfail over this test with a link to the issue here in a code comment, so that we can remember to fix this. Or this can stay as an open PR until it's fixed; however you like.

ryanking13:
I am pretty sure that we cannot reproduce the error with this, as `Py_FinalizeEx` is not called in our test suite.

greateggsgreg:
Ah, you're right, this will only fail if we run via non-prod CLI. You've proposed fixes for the underlying cause (that's for that) elsewhere, so, this change is now just a harmless set of jsonschema tests

ryanking13:
@greateggsgreg Let me close this as adding tests doesn't actually add much value I think. Let's fix the finalize issue in upstream instead.

PR #582, opened 2026-05-15
Disable pplpy

ryanking13:
## Description

~~The pplpy version we have seems to have issue with building python 3.14. Trying to update it to a newer version and see if it works~~

Disabling pplpy for now to make the CI pass

cc: @mkoeppe 

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [ ] Other (please describe)

### DIFF
--- a/packages/pplpy/meta.yaml
+++ b/packages/pplpy/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: pplpy
-  version: 0.8.10
+  version: 0.9.0
   top-level:
     - ppl
 source:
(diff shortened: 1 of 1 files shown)


ryanking13:
Looks like it doesn't work 😢 .

@mkoeppe Let me disable pplpy for now to move our build forward. You can re-enable it or maybe start building and publishing wheels out-of-tree as PyPI now supports uploading pyodide/pyemscripten wheels.

mkoeppe:
I'd suggest switching from pplpy to the API-compatible fork https://pypi.org/project/passagemath-ppl/, which is properly maintained

PR #591, opened 2026-06-06
Set submodule to pyodide/pyodide-build#21

agriyakhetarpal:
## Description

A PR to test https://github.com/pyodide/pyodide-build/pull/21

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [x] Other (please describe): testing whether f2py and numpy-config executables installed as console scripts from cross-build packages can now be found on PATH for SciPy

---

> [!NOTE]
>
>
>


agriyakhetarpal:
The `nlopt` build failure is slightly related but also unrelated to our changes. It only has a `setup.py` file and no declarative metadata, so it's passing on to the legacy setuptools build backend.

Edit: moved to #592 and merged main. Resolving!

agriyakhetarpal:
This unfortunately doesn't work well :(

The problem with RobotRaconteur is that it needs access to `ndarrayobject.h`, which was present when we had `PYTHONPATH` pointing at a usable cross-built NumPy but no longer do.

I don't think it would be correct for us to list the entire `numpy/_core/include/numpy/` directory because it contains 20 header files, such as `arrayobject.h`, `npy_common.h`, `ufuncobject.h`, etc., and then the `libdivide/` and `random/` stuff is there too. That is going to be a big maintenance burden for us, because those are very fragile and will keep changing a lot. It also kind of defeats the purpose of the `cross-build-files:` field, which is meant to contain platform-specific files, since these headers are largely independent of the platform.

Since we can't really change RobotRaconteur's recipe here to build the wheel within its CMake scripts (too big of a change), I think it would be better to restore a mechanism that keeps `$HOSTINSTALLDIR/.../numpy/` populated with a WASM-patched `numpy` installation. So we probably need to keep the `old pip install --no-deps -t host_site_packages` step (or an equivalent) _in addition to_ the new "install into isolated build venv" mechanism.

Another reason is that OpenCV also needs access to this header and will currently fail. Other C/C++ packages might need it too, for non-pypa/build based builds.

agriyakhetarpal:
All green now!

agriyakhetarpal:
Closing this PR as a rough draft, will move the changes into another one.

PR #593, opened 2026-06-07
Download CLAPACK from Internet Archive for libf2c

agriyakhetarpal:
## Description

We often come across errors in our CI, of the form

```python
RuntimeError: Failed to download
http://www.netlib.org/clapack/clapack.tgz after 3 trials
```

This happens a lot with GitHub Actions runners. Let's try using the Internet Archive for such links; hopefully it will be more reliable. I confirmed that the SHA-256 checksum is correct and matches the current one.

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [x] Other (please describe)

---

> [!NOTE]
>
>
>


agriyakhetarpal:
Merging to see if this works fine for subsequent builds (I hope it does)

PR #598, opened 2026-06-08
Remove RobotRaconteur and libopenssl from the distribution

agriyakhetarpal:
## Description

This PR removes RobotRaconteur from the distribution, following @johnwason's advice in https://github.com/pyodide/pyodide-build/issues/365#issuecomment-4645548837 (and cc: @ryanking13). RobotRaconteur is one of the first few packages to start publishing PyEmscripten wheels to PyPI following the acceptance and adoption of PEP 783.

I have also removed the OpenSSL recipe here, as RobotRaconteur was the only package that was using it. However, I recognise that OpenSSL is kind of an important package and maybe we should keep it, but I think it's pretty easy for us to bring it back from the Git history if some package needs it at a later point.

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [x] Other (please describe): Removing a package

---


ryanking13 (approved):
Sounds good. For libopenssl, yeah we can restore it anytime if needed.

PR #602, opened 2026-06-12
Remove imgui-bundle from the distribution

pthom:
imgui-bundle now publishes PEP 783 `pyemscripten` wheels on PyPI (starting with [1.92.801](https://pypi.org/project/imgui-bundle/1.92.801/#files): `cp314-cp314-pyemscripten_2026_0_wasm32`), so micropip resolves it directly from PyPI and the recipe is no longer needed. 

I am the maintainer of imgui-bundle (and of this recipe).

PS: many thanks for your assistance in publishing the initials wheels here


agriyakhetarpal (approved):
Thank you very much for upstreaming the pyemscripten wheel builds to PyPI for imgui-bundle!

PR #605, opened 2026-06-23
Remove blosc2, gsw, and pycdfpp recipes

ryanking13:
Removing these as all these wheels have pyemscripten wheels in PyPI now (#599).

cc: @jeandet and @FrancescAlted for visibility


agriyakhetarpal (approved):

FrancescAlted:
Hey, I am getting issues when trying to load the latest blosc2 wheel in PyPI when tried to be used in jupyterlite.

In particular, this:

```
    # The blosc2 >= floor is important: Pyodide bundles its own (often older) blosc2
    # in the distribution lock (e.g. 4.1.2 on 314.0.0), and micropip prefers a
    # bundled package over PyPI unless the requirement excludes it. The constraint
    # forces micropip to fetch a current release from PyPI instead of the stale
    # bundled one.
    await micropip.install("blosc2>=4.6.0")
    import blosc2
    await micropip.install("caterva2")
    import caterva2
    print(f"Installed blosc2 {blosc2.__version__} and caterva2 {caterva2.__version__} successfully!")
```

is failing with:

```
File /lib/python3.14/site-packages/micropip/transaction.py:139, in Transaction.check_version_satisfied(self, req, allow_reinstall)
    137     return False, ""
    138 else:
--> 139     raise ValueError(
    140         f"Requested '{req}', "
    141         f"but {req.name}=={ver} is already installed. "
    142         "If you want to reinstall the package with a different version, "
    143         "use micropip.install(..., reinstall=True) to force reinstall, "
    144         "or micropip.uninstall(...) to uninstall the package first."
    145     )

ValueError: Requested 'blosc2>=4.6.0', but blosc2==4.1.2 is already installed. If you want to reinstall the package with a different version, use micropip.install(..., reinstall=True) to force reinstall, or micropip.uninstall(...) to uninstall the package first.
```

So I think this "but blosc2==4.1.2 is already installed" is because pyodide installs it first.  I tried using `micropip.install(..., reinstall=True)`, but there is some kind of weird interaction with 4.1.2.  May I suppose that the next pyodide release will remove blosc2 completely and the above would work?  If so, is there any expected timeframe for the release?

Thanks!

hoodmane:
> May I suppose that the next pyodide release will remove blosc2 completely

As soon as this commit is taken up yes.

> If so, is there any expected timeframe for the release?

I think 314.0.1 already went out with it. 

> Jupyterlite 

I think is still on Pyodide 0.29? So I think your problem is downstream. What is `pyodide.__version__`?

agriyakhetarpal:
JupyterLite is already on Pyodide 314.0.1. I updated to 314.0.0 in https://github.com/jupyterlite/pyodide-kernel/pull/294, which landed in v0.8.0, and there was one more change after that to update to 314.0.1 which landed in v0.8.1: https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.8.1. JupyterLite v0.8.0 should work with the Pyodide kernel v0.8.1.

ryanking13:
> So I think this "but blosc2==4.1.2 is already installed" is because pyodide installs it first. I tried using micropip.install(..., reinstall=True), but there is some kind of weird interaction with 4.1.2. May I suppose that the next pyodide release will remove blosc2 completely and the above would work? If so, is there any expected timeframe for the release?

I cannot reproduce the same error in Pyodide 314.0.1 with your example. Jupyterlite applies some patches against micropip so it can be a jupyterlite issue.

hoodmane:
This is at least on Pyodide 0.29.0:
https://jupyter.org/try-jupyter/lab/?mode=multiple-document

agriyakhetarpal:
That one is at https://github.com/jupyter/try-jupyter, which isn't updated as frequently. https://jupyterlite.readthedocs.io/en/stable/try/lab would be better.

That said, it's being updated: https://github.com/jupyter/try-jupyter/pull/83

FrancescAlted:
Yes, with https://jupyterlite.readthedocs.io/en/stable/_static/lab/index.html I can reproduce the issue:

```
import micropip
await micropip.install(["blosc2>=4.6.0"])
import blosc2
print(f"Installed blosc2 {blosc2.__version__}")
```

gives:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[1], line 2
      1 import micropip
----> 2 await micropip.install(["blosc2>=4.6.0"])
      3 import blosc2
      4 print(f"Installed blosc2 {blosc2.__version__}")

File /lib/python3.13/site-packages/micropip/package_manager.py:202, in PackageManager.install(self, requirements, keep_going, deps, credentials, pre, index_urls, constraints, reinstall, verbose)
    187 wheel_base = Path(getsitepackages()[0])
    189 transaction = Transaction(
    190     _compat_layer=self.compat_layer,
    191     ctx=ctx,  # type: ignore[arg-type]
   (...)    200     reinstall=reinstall,
    201 )
--> 202 await transaction.gather_requirements(requirements)
    204 if transaction.failed:
    205     failed_requirements = ", ".join(
    206         [f"'{req}'" for req in transaction.failed]
    207     )

File /lib/python3.13/site-packages/micropip/transaction.py:74, in Transaction.gather_requirements(self, requirements)
     66 async def gather_requirements(
     67     self,
     68     requirements: list[str] | list[Requirement],
     69 ) -> None:
     70     requirement_promises = [
     71         self.add_requirement(requirement) for requirement in requirements
     72     ]
---> 74     await asyncio.gather(*requirement_promises)

File /lib/python3.13/site-packages/micropip/transaction.py:90, in Transaction.add_requirement(self, req)
     83     if as_req.name.endswith(".whl"):
     84         # This happens when the requirement is passed as a relative URL:
     85         # For instance, micropip.install("pkg-1.0.0-py3-none-any.whl")
     86         # packaging cannot distinguish this is a relative URL or a very weird wheel name ... sigh ...
     87         # But we want to treat it as a custom download location.
     88         return await self.add_requirement_from_url(req)
---> 90     return await self.add_requirement_inner(as_req)
     91 except InvalidRequirement:
     92     if not urlparse(req).path.endswith(".whl"):

File /lib/python3.13/site-packages/micropip/transaction.py:202, in Transaction.add_requirement_inner(self, req)
    199 # Is some version of this package is already installed?
    200 req.name = canonicalize_name(req.name)
--> 202 satisfied, ver = self.check_version_satisfied(
    203     req, allow_reinstall=self.reinstall
    204 )
    206 if satisfied:
    207     logger.info("Requirement already satisfied: %s (%s)", req, ver)

File /lib/python3.13/site-packages/micropip/transaction.py:139, in Transaction.check_version_satisfied(self, req, allow_reinstall)
    137     return False, ""
    138 else:
--> 139     raise ValueError(
    140         f"Requested '{req}', "
    141         f"but {req.name}=={ver} is already installed. "
    142         "If you want to reinstall the package with a different version, "
    143         "use micropip.install(..., reinstall=True) to force reinstall, "
    144         "or micropip.uninstall(...) to uninstall the package first."
    145     )

ValueError: Requested 'blosc2>=4.6.0', but blosc2==3.5.1 is already installed. If you want to reinstall the package with a different version, use micropip.install(..., reinstall=True) to force reinstall, or micropip.uninstall(...) to uninstall the package first.
```

Our site can also be used for reproducing that at e.g.: https://cat2.cloud/demo/roots/@public/examples/mandel-jit-vs-nojit.ipynb, with the difference that "blosc2==4.1.2 is already installed" is issued instead of "blosc2==3.5.1 is already installed" above (maybe because we are using JupyterLite 0.8.0?).

And you are right, Pyodide does not seem to be the issue, because https://blosc.org/demos/newton-js-vs-numpy-vs-nojit-vs-jit.html can run the blosc2 4.6.0 wheel without issues.

hoodmane:
Try moving the import to blosc into a separate cell after micropip.install()?

agriyakhetarpal:
Correct, the `import blosc2` line in the Pyodide kernel has special behaviour to look up the package namespace-to-name mapping in the lockfile and install whatever version is present there. If you do `micropip.install()` in one cell and `import blosc2` in the next cell, it should work fine.

FrancescAlted:
> Try moving the import to blosc into a separate cell after micropip.install()?

Hey, that works beautifully for our platform!  Although this looks like magic to me: would you mind explaining why the cell split works?

Thanks a lot!

agriyakhetarpal:
Great, nice that it's resolved! This is the logic for the special behaviour to install what is being `import`ed: https://github.com/jupyterlite/pyodide-kernel/blob/823e2772da097fa6f6e36aebeac1545a44ceb337/packages/pyodide-kernel/py/pyodide-kernel/pyodide_kernel/kernel.py#L101-L118

Essentially, for every cell, we hand the code to Pyodide's `loadPackagesFromImports`, which parses the `import` statements, looks up each import name in the lockfile's import-name to package-name mapping (we provide an `imports` field of each package entry in `pyodide-lock.json`).

For arbitrary packages to be installed via `%pip install` magics, they are handed off to `piplite` and follow a slightly different code path, ultimately reaching `micropip` as `await piplite.install("somepkg")`. `piplite` is a wrapper over `micropip`.

PR #606, opened 2026-06-23
Remove fastcan from the distribution

MatthewSZhang:
## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [x] Other (please describe)

`fastcan` from v0.6.0 has pyemscripten wheels (i.e., `fastcan-0.6.0-cp314-cp314-pyemscripten_2026_0_wasm32.whl` and `fastcan-0.6.0-cp313-cp313-pyemscripten_2025_0_wasm32.whl`) on [pypi](https://pypi.org/project/fastcan/#files), so micropip resolves it directly from PyPI and the recipe is no longer needed.

---

> [!NOTE]
>
>
>


agriyakhetarpal (approved):
Thanks for removing this recipe, @MatthewSZhang! I also checked your project's CI, and everything seems to be in order. pyemscripten wheels are available on https://pypi.org/project/fastcan/0.6.0/#files, too.

Note that with pypa/cibuildwheel#2609 (merged but not released – we'll probably add it to v4.2.0), `pyodide-build` will be the only supported build frontend for the pyodide platform and `build[uv]` will not work: https://github.com/scikit-learn-contrib/fastcan/blob/4cd6b2f6b6f692d10655aa5f518a69da120c983b/.github/workflows/emscripten.yml#L15

Issue #604, opened 2026-06-22
Progress on a SciPy that is free from a Fortran-to-WASM compiler dependency and upstreaming WASM patches


agriyakhetarpal:
At the time of writing, SciPy's `main` branch has dropped `scipy.odr`, which was already deprecated and was the last piece of Fortran code present in SciPy. SciPy v1.19.0, expected to arrive in December 2026/January 2027, will be the first version of a Fortran-free SciPy. Note that it was already possible to build SciPy v1.18.0 with `scipy.odr` excluded via the `without_fortran` Meson option, which I had tried locally and succeeded with, but didn't end up including it in #586 as we ought to match what the SciPy v1.18 PyPI wheels ship to users.

Building on top of #586 locally in a separate branch and excluding `scipy.odr` from the build, I have recently been at work compiling SciPy to WebAssembly and establishing a recipe that removes most of the f2c-related and SciPy-related hacks in pyodide-build, drops the use of a Fortran compiler and of libf2c across the stack:

I posted about this in this comment in the Scientific Python Discourse forum last week: https://discuss.scientific-python.org/t/ann-scipy-1-18-0-release/2394/2, and I will use this issue to post a longer comment about the current progress I have:
- While OpenBLAS enables a `NO_FORTRAN` build option to only build the BLAS parts and get the LAPACK parts from a different LAPACK library, I found trouble using it for our needs, because it's written in a manner that still relies on Fortran-like calling conventions, even though we don't compile any Fortran-specific parts with its C codebase. This makes upstreaming tricky because of the large number of function signatures that would not match (see below).
- As a result, I started looking for other BLAS libraries that do not rely on Fortran, and I arrived at [BLIS](https://github.com/flame/blis). It looked promising to me because it is entirely written in C.
  - I was able to compile it to WASM, albeit with a few differences, as it requires https://github.com/flame/blis/issues/916, which is being added by https://github.com/flame/blis/pull/926 (which I'm using as a patch). I hit this issue when running the SciPy test suite, where I noticed null function signature mismatches where these symbols representing Level 1 plane rotation routines were not exported.
  - Also, I am currently using the generic architecture/kernel for BLIS, instead of putting in the effort to write an optimised wasm32 kernel.
- In all this, I also fixed an issue with RPATHs showing up in macOS's `/private/` folders and SciPy extension modules being non-relocatable (https://github.com/pyodide/auditwheel-emscripten/pull/55). 
[...]

PR #609, opened 2026-06-30
Add a recipe for the BLIS framework and associated patches

agriyakhetarpal:
## Description

This adds a recipe for [BLIS](https://github.com/flame/blis) [v2.1](https://github.com/flame/blis/releases/tag/2.1), a BSD-licensed modern alternative to the traditional BLAS flavours designed to instantiate high-performance dense linear algebra libraries.

The benefit of the library is that it is all C99 and has no Fortran bits, so it compiles cleanly with Emscripten with these changes. It uses architecture-specific microkernels that are written in Assembly for performance reasons. However, obviously, is no kernel for WASM here, so we have no performance benefits of the microkernel architecture, and thus I've compiled it using only a generic target.

I am also experimenting with compiling with SIMD intrinsics locally with the use of the `-msimd128` flag, which works, but I need to benchmark it properly. I'll split that into a separate PR.

I plan to use this recipe further for SciPy's BLAS needs. This is not the final state the recipe will be in, as it needs some further changes (which I've intentionally not included here). 

See #604
xref flame/blis#491
from https://github.com/agriyakhetarpal/blis/pull/1
from https://github.com/agriyakhetarpal/blis/pull/2
See also https://github.com/flame/blis/issues/916 and https://github.com/flame/blis/pull/926

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [x] Other (please describe): Adding a new library, to be later used with consuming packages

---

> [!NOTE]
>
>
>

### DIFF
--- a/packages/libblis/meta.yaml
+++ b/packages/libblis/meta.yaml
@@ -1,0 +1,42 @@
+package:
+  name: libblis
+  version: "2.0"
+  tag:
+    - library
+    - shared_library
+source:
+  url: https://github.com/flame/blis/archive/refs/tags/2.1.tar.gz
+  sha256: 901752ec596cd63421ea45a6a06a9f34491cf5ae01cea412ece05f97a2690a96
+  patches:
+    # Derived from https://github.com/agriyakhetarpal/blis/pull/1
+    - patches/0001-Compile-BLIS-to-WebAssembly.patch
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
All green! I am merging this for now to further speed up my iterations, but please let me know if you spot any issues.

PR #613, opened 2026-07-01
Update OpenBLAS to 0.3.31

teddygood:
## Description

This updates the `libopenblas` recipe from OpenBLAS 0.3.28 to 0.3.31, matching the OpenBLAS line currently used by upstream SciPy. See #604. We plan to update to OpenBLAS 0.3.32/0.3.33 once upstream SciPy starts using those versions in a released version.

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [ ] Other (please describe)

---

> [!NOTE]
>
>
>

### DIFF
--- a/packages/libopenblas/meta.yaml
+++ b/packages/libopenblas/meta.yaml
@@ -1,13 +1,13 @@
 package:
   name: libopenblas
-  version: 0.3.28
+  version: 0.3.31
   tag:
     - core
     - library
     - shared_library
 source:
-  sha256: f1003466ad074e9b0c8d421a204121100b0751c96fc6fcf3d1456bd12f8a00a1
-  url: https://github.com/OpenMathLib/OpenBLAS/releases/download/v0.3.28/OpenBLAS-0.3.28.tar.gz
+  sha256: 6dd2a63ac9d32643b7cc636eab57bf4e57d0ed1fff926dfbc5d3d97f2d2be3a6
+  url: https://github.com/OpenMathLib/OpenBLAS/releases/download/v0.3.31/OpenBLAS-0.3.31.tar.gz
   patches:
     - patches/0001-Add-Wno-return-type-flag.patch
     - patches/0002-Align-xerbla_array-signature-with-scipy-expectation.patch
--- a/packages/libopenblas/meta.yaml
+++ b/packages/libopenblas/meta.yaml
@@ -26,6 +26,7 @@ build:
     sed -ri 's/void(\s+)(C?NAME)/int\1\2/g' interface/*.c
     sed -ri 's/((extern)?.+) void ([a-z0-9]+_)/\1\2 int \3/g' lapack-netlib/SRC/*.c \
         lapack-netlib/SRC/DEPRECATED/*.c
+    sed -ri 's/^void (LAED4|LACPY|LASET)\(/int \1(/g' lapack/laed3/*.c
     # For some functions (mostly handling complex I think) f2c actually
     # generate a function that returns void so I need to revert the void to int
     # change the previous line does.
--- a/packages/libopenblas/patches/0003-Skip-linktest.patch
+++ b/packages/libopenblas/patches/0003-Skip-linktest.patch
@@ -11,13 +11,13 @@ diff --git a/exports/Makefile b/exports/Makefile
 index 7682f851d..86b2cd2f4 100644
 --- a/exports/Makefile
 +++ b/exports/Makefile
-@@ -184,24 +184,24 @@ ifeq ($(F_COMPILER), INTEL)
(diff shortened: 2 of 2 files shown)


agriyakhetarpal:
Thanks @teddygood! I just added the "scipy" label to launch the SciPy tests as part of this PR. We should be good to go once the CI passes 🤞🏻

agriyakhetarpal (commented):
Oops, a crash in `test_optimize.py`... unfortunately, pytest is a bit too verbose when we run it with the `-v` option, so we disable it in CI. As a result, it's not easy to see what specific test it is crashing on. While the setup to test SciPy locally is a bit of a mess, you should be able to follow the exact steps as in the CI job after building SciPy using `build-recipes`, and that will get you to the same SciPy test suite up and running. `RuntimeError: unreachable` suggests that this isn't a signature mismatch issue, but more of a test-related problem. At least we have some idea from the traceback that this is occurring in `scipy.linalg.eigh` and via `test_MikotaPair`: https://github.com/scipy/scipy/blob/09bac46b69447b454b775eb2793f9fb848e4ec9f/scipy/sparse/linalg/_eigen/lobpcg/tests/test_lobpcg.py#L575-L614, but we don't know which case and why.

Some useful things that I like doing, specific to the context within working on SciPy:
- looking at the test upstream, i.e., whether it is skipped for a particular architecture or has a test input/parametrization that would be absurd in WASM. (Edit: probably ruled out, by looking at the test?)
- any upstream failures with OpenBLAS versions reported in the SciPy issue tracker that may be related, even though they may not be specific to this test
- building using the debug builds. You can use `pyodide xbuildenv search --debug` and install a debug xbuildenv, which is available with the 314.0.2 release: https://github.com/pyodide/pyodide/releases/tag/314.0.2. `pyodide xbuildenv install --debug 314.0.2` should install it. This will also provide a debug variant of the Pyodide Python interpreter, which prints out quite a lot, but is still useful.
- Updating to OpenBLAS 0.3.29 and 0.3.30 first and seeing if it reproduces against that (you can also introduce a matrix in the CI job temporarily to build SciPy against OpenBLAS versions and tests in parallel, if that is helpful somewhat – and Claude Code or similar can help set that up pretty quickly)

Let's hope this won't be too gnarly to fix – but it surely is tedious!

teddygood:
> Some useful things that I like doing, specific to the context within working on SciPy:
> 
> * looking at the test upstream, i.e., whether it is skipped for a particular architecture or has a test input/parametrization that would be absurd in WASM. (Edit: probably ruled out, by looking at the test?)
> * any upstream failures with OpenBLAS versions reported in the SciPy issue tracker that may be related, even though they may not be specific to this test
> * building using the debug builds. You can use `pyodide xbuildenv search --debug` and install a debug xbuildenv, which is available with the 314.0.2 release: https://github.com/pyodide/pyodide/releases/tag/314.0.2. `pyodide xbuildenv install --debug 314.0.2` should install it. This will also provide a debug variant of the Pyodide Python interpreter, which prints out quite a lot, but is still useful.
> * Updating to OpenBLAS 0.3.29 and 0.3.30 first and seeing if it reproduces against that (you can also introduce a matrix in the CI job temporarily to build SciPy against OpenBLAS versions and tests in parallel, if that is helpful somewhat – and Claude Code or similar can help set that up pretty quickly)
> 
> Let's hope this won't be too gnarly to fix – but it surely is tedious!

Thanks for the detailed pointers! They were very helpful for narrowing this down. I followed the CI setup locally and tried the checks you suggested. The crash is coming from `test_MikotaPair`, specifically through the `scipy.linalg.eigh(..., driver="gvd")` path. I was able to reduce it further to the `DSTEVD -> DSTEDC -> DLAED*` path: smaller cases pass, but the crash starts once the divide-and-conquer LAPACK path is used.

I also tried building OpenBLAS 0.3.29 and 0.3.30 as suggested. Both versions pass the same focused SciPy checks, and neither of them has the new `lapack/laed3` directory. In the 0.3.31 build log, I found `wasm-ld: warning: function signature mismatch: dlaed4_`, which pointed at the new `lapack/laed3/laed3_single.c` path added in 0.3.31.

After aligning the f2c-style return signatures in the new `lapack/laed3/*.c` sources, the focused 0.3.31 repro passes:

- `scipy.linalg.eigh(..., driver="gvd")` focused probe passes
- `scipy.sparse.linalg._eigen.lobpcg.tests.test_lobpcg::test_MikotaPair`: `7 passed`
- full `test_lobpcg.py`: `1540 passed, 1 skipped`

So I think this was a signature mismatch after all, but in the new 0.3.31 `laed3` path rather than in the more obvious BLAS/LAPACK entry point. Hopefully the CI passes this time.

agriyakhetarpal (approved):
Nice work investigating this, thank you, @teddygood!

PR #614, opened 2026-07-01
upgreade to version 14, then version 15

airen1986:
## Type of change

- [ ] Adding a new package
- [X] Updating an existing package
- [ ] Bug fix
- [ ] Other (please describe)

---

> [!NOTE]
>
>
>

Upgrading highspy to v14 and then version 15

### DIFF
--- a/packages/highspy/meta.yaml
+++ b/packages/highspy/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: highspy
-  version: 1.13.1
+  version: 1.14.0
   top-level:
     - highspy
 source:
-  url: https://files.pythonhosted.org/packages/source/h/highspy/highspy-1.13.1.tar.gz
+  url: https://files.pythonhosted.org/packages/source/h/highspy/highspy-1.14.0.tar.gz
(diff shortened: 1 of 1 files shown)


ryanking13:
Hi @airen1986! Could you please try to upstream this package to `highspy` repository? As you can see in the PR template, we are not actively accepting new contributions in this repository unless there is a clear reason.

airen1986:
yes, I would do that, I am seeing some issues in cibuildwheel for v15, which is working well in v14. I created this PR so we can see what issue that is and then you can advise how I could resolve that

airen1986:
@ryanking13 till v14 it was just producing on shared object _core.so, now it is generating 2 shared objects _core.so and libhighs.so, so I am seeing an error like this:
Failed to load dynamic library
.../highspy/_core.so

Failed to find dynamic library "libhighs.so"

airen1986:
from chatGPT:
This means:
_core.so itself loaded correctly.
While loading, the dynamic loader discovered that _core.so has a dependency on libhighs.so.
Pyodide couldn't find libhighs.so in its virtual filesystem as it doesn't have a system library path like Linux.

PR #618, opened 2026-07-07
Remove apsw, awkward-cpp, flint

ryanking13:
These packages now have 2026-pyemscripten wheel (#599) and can be installed from PyPI directly, so removing recipes.

FYI @airen1986, @agoose77, @mkoeppe, @oscarbenjamin


agriyakhetarpal (approved):

ryanking13:
flint was taking more than 50 minutes (libflint + python-flint) to build, so this would give us a lot of improvements in CI time.

PR #586, opened 2026-05-26
SciPy v1.18.0

agriyakhetarpal:
## Description

This PR updates our SciPy recipe to version 1.18.0, released today: https://github.com/scipy/scipy/releases/tag/v1.18.0.

- Several patches have been dropped, and we are now down from eleven to six. This can become five once we figure out a way to ensure `numpy-config` is not missing from PATH, which is the current basis for patch 0006. That might be resolved via https://github.com/pyodide/pyodide-build/pull/21 at some point. I'll get to that PR and try out some things next.
- I was able to figure out a way to bring back the test summaries and warnings report without invoking the previous fatal error (the function signature mismatch occurring due to C++ global destructors). I'm not sure, but I think some of it has also been happening due to https://github.com/scipy/scipy/issues/25208, but since we see `<no Python frame>` in the traceback, it doesn't suggest it. We should investigate at some point – there is no action here.
- Tons of xfails related to FORTRAN77 code are now removed, and some numerical differences have also beautifully coalesced into nothing. Almost all of the test failures/skips are related to threading-related tests, though some xfails do remain, of course.
- I have **not** used the new `-Dwithout_fortran` build option here just yet, as `scipy.odr` is deprecated but not removed. I intend to leave it for another PR.

For more information, please refer to https://discuss.scientific-python.org/t/proposed-release-schedule-for-scipy-1-18-0/2308.

ccing the recipe maintainers: @lesteve and @steppi, and also @rgommers; for visibility.

- Closes pyodide/pyodide#5060
- Closes pyodide/pyodide#3380
- Closes pyodide/pyodide#5211
- Closes https://github.com/pyodide/pyodide/issues/4249 (it's a bit out of date, and the MRE in https://github.com/pyodide/pyodide/issues/4249#issuecomment-1808328455 no longer reproduces)

#### Notes

- See reproducer/test code for pyodide/pyodide#5211 below:

	<details><summary>Click to expand</summary>
	<p>
	
	```pycon
	>>> import ctypes
	>>> import scipy.linalg.cython_lapack as cl
	>>> import scipy.linalg.cython_blas as cb
	>>> ctypes.pythonapi.PyCapsule_GetName.restype = ctypes.c_char_p
	>>> def capsule_sig(cap):
	...     return ctypes.pythonapi.PyCapsule_GetName(ctypes.py_object(cap)).decode()
	... 
	>>> dgesv_sig = capsule_sig(cl.__pyx_capi__['dgesv'])
[...]

PR #619, opened 2026-07-08
Compile a Fortran-free WASM SciPy, with BLIS and semicolon-lapack

### LINKED ISSUE #604: Progress on a SciPy that is free from a Fortran-to-WASM compiler dependency and upstreaming WASM patches

agriyakhetarpal:
## Description

See #604 for all the context. This is the last stage of work towards a rendition of SciPy that builds without a Fortran-to-WASM compiler, weird f2c hacks, Fortran calling conventions, function signature mismatches, and general runtime issues that have troubled us over the years. With the changes in this PR and https://github.com/pyodide/pyodide-build/pull/395 in place (used in the submodule in this PR), the stack is entirely based in C, with no Fortran code. OpenBLAS is no longer used with SciPy, as was once integrated with https://github.com/pyodide/pyodide/pull/3331. The stack is now based on https://github.com/flame/blis (see https://github.com/agriyakhetarpal/blis/pull/1) and https://github.com/ilayn/semicolon-lapack.

It is important to note that this is not ready for merging yet, as `scipy.odr` is still present in SciPy 1.18, and we should keep it. We should wait until SciPy 2.0 later this year to do so, as that subpackage will have been removed in that update (it has already been removed in SciPy `main`). However, what we have achieved through this effort now paves the way forward for an in-tree CI job and subsequent first-class support for SciPy in Pyodide/WebAssembly.

Note that we still need one patch for SciPy (but we are down from five patches to just one!) and a few for BLIS. I upstreamed the patches semicolon-lapack needed in https://github.com/ilayn/semicolon-lapack/pull/23 and https://github.com/ilayn/semicolon-lapack/pull/24). A 100% of the SciPy test suite is passing in WebAssembly as far as can be made applicable (see `scipy-conftest.py`). There is one skipped test, but that's more of a test ordering problem, and the test definitely passes in isolation.

This PR follows on from the last updates to the SciPy recipe that I made in #586 and #611.

cc: the listed recipe maintainers @steppi and @lesteve, and also @ilayn and @rgommers. Thank you!

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [ ] Other (please describe)

---

> [!NOTE]
>
>
>


agriyakhetarpal:
The entirety of the test suite is passing in https://github.com/pyodide/pyodide-recipes/actions/runs/28972682213/job/85976430414?pr=619 🎉

(Looks like adding the "do not merge" label triggered the test suite again, oh well, but never mind :P)

My next steps here will be to:
- Upstream the SciPy getrf/getri function signature mismatch (I'm preparing a PR)
- Start figuring out what to do about the `gees_` calls patch 0003 – @ilayn, I think you know much better than I do about it!
- Wait for a new semicolon-lapack tag and bring in https://github.com/ilayn/semicolon-lapack/pull/24.

The BLIS patches are also somewhat upstreamable. I think it would be nice if we were to host a GitHub release somewhere with a precompiled, patched WASM BLIS build and a semicolon-lapack build that SciPy can use for a CI job. Or, it's useful to have these as submodules or similar under the SciPy org, as the compilation time isn't a lot for either of them and that helps avoid the churn of having to create a new release for every code change, as code can be patched pretty easily. The only concern with a SciPy WASM CI job, IMO, would be how fragile it would be, but that's (i) something we can only tell when we try it, and (ii) we cannot escape that reality (though we can easily disable the job in the Actions tab or xfail tests temporarily if there are issues).

Overall, I think https://github.com/ilayn/semicolon-lapack/pull/22 helped quite a lot here. I haven't really bisected things as such, but I believe I got the numerical issues I was getting fixed for free from that PR, in addition to the test suite crash at 57%. Previously, there were thousands of tests failing due to irreconcilable and absurd values (such as 3.5324234 instead of 2.108471 and so on), but there are none now!

ilayn:
Super cool that you pulled it off @agriyakhetarpal I just made a dummy release again just to keep things rolling. 

For the build process, I think if you skip building the tests (`-Dtests=false` should do it), it is bearable times to build it from scratch. Not sure about the download times though. So indeed having a prebuilt binary somewhere might be a good idea.

I'll check the patch no. 3.

rgommers:
`libblis` + `libsemilapack` in your build time table is ~6 minutes together, if that's on regular GitHub runners then that's acceptable for a SciPy CI job I'd say.

agriyakhetarpal:
Yup, that's also when all packages are being built on an oversubscribed CPU, too. So building them with parallel jobs sequentially should definitely reduce that time slightly down from 6 minutes.

packages/libblis/patches/0003-WASM-ABI-void-returns-lsame-xerbla_array.patch
ryanking13, line 1: Are these patches upstreamed?

packages/scipy/meta.yaml
```diff
@@ -133,15 +85,11 @@ build:
...
   host:
     - numpy
-    - libopenblas
+    - libblis
+    - libsemilapack
     - libboost
   run:
     - numpy
-  executable:
-    - gfortran
```
ryanking13, line 141: yeeees

environment.yml
```diff
@@ -9,6 +9,8 @@ dependencies:
   - swig
   - make
   - cmake<4
+  - meson
```
ryanking13, line 12: why do we need host meson? doesn't scipy install it in the isolated environment?

ryanking13 (approved):
I guess we will need to think about when this should be merged (when the f2c, scipy-related codes can be dropped from pyodide-build), but amazing work!

packages/libblis/patches/0003-WASM-ABI-void-returns-lsame-xerbla_array.patch
agriyakhetarpal: Nope, not yet, but they live in https://github.com/pyodide/blis. I plan to do this soon-ish and send PRs to BLIS, but they span quite a lot of files, so I'll take it slowly :) I'm waiting for a response from the BLIS maintainers on https://github.com/flame/blis/issues/491 first.

environment.yml
agriyakhetarpal: This is needed for semicolon-lapack, actually, which uses Meson as its build system. It's just that we haven't had any system-level library recipes so far that were using Meson, and this is the first.

If it were a Python package, then yes, it would have been picked up as a build-time dependency.

ryanking13: I see. Thanks for the clarafication!

agriyakhetarpal:
The build passes with Emscripten 6.0.5, but it looks like it's still building 2026_0 wheels when we want it to build with Python 3.15 – I see `Processing ./dist/scipy-1.18.0-cp314-cp314-pyemscripten_2026_0_wasm32.whl` in the logs. Is this something that should signal trouble with our overall testing setup and we should be building with Python 3.15 in #639 instead of Python 3.14?

PR #621, opened 2026-07-13
Upgrade numpy to v2.4.6

juntyr:
## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [ ] Other (please describe)

---

> [!NOTE]
>
>
>


agriyakhetarpal (approved):
Thanks!

PR #639, opened 2026-07-31
Emcc 6.0.5

hoodmane:


agriyakhetarpal:
I am kinda surprised that the "full build" label doesn't run the full SciPy test suite, given that the test suite often exposes problems (but maybe we're lucky here and there won't be any). I'll change that in another PR.

PR #646, opened 2026-08-09
Fix zfpy top level packages

juntyr:
## Description

zfpy includes top level entries for `cuda_zfp`, `share`, `template`, and `zfpy` - only the last is the real extension module, the others are mistakenly auto-discovered but are just the names of C source folders in libzfp. The newly added patch ensures that only `zfpy` is included in the top level entries.

## Type of change

- [ ] Adding a new package
- [x] Bug fix
- [ ] Other (please describe)

---

> [!NOTE]
>
>
>


ryanking13:
Could you elaborate on what does this exactly fix? Could you add test cases? I don't understand what

> mistakenly auto-discovered

means.

juntyr:
This recipe uses setuptools, which automatically tries to find top-level packages. Because the src/ layout is common, setuptools looks at the libzfp source structure, which has some files in src/cuda_zfp, src/share, and src/template, and thinks that those will be Python packages, in addition to the zfpy Python extension. But they have nothing to do with Python, and so the wheel’s top level meta file contains unnecessary entries. By turning off auto package discovery, we only get the correct top level.

ryanking13:
I see. Makes sense. Personally, I don't use setuptools's top_level.txt file at all (I think it is useless and rather creates more problem 🤷), but having this patch wouldn't harm anything.

Could you please add a comment on the patch file so users who later see that patch files do not have to look up this PR? Otherwise looks good to me.

ryanking13 (approved):

PR #647, opened 2026-08-13
Add database driver packages

ryanking13:
## Description

This adds `mysqlclient` and `psycopg` package which are two famous, non-pure database driver packages. Django is using these packages by default when connecting to mysql or postgresql so I would like to add support for these packages.

As you can see, the build step is non-trivial (did a lot of back and forth with LLMs). Ideally, we should upstream these packages, but I would like to land here first.

## Type of change

- [x] Adding a new package
- [ ] Bug fix
- [ ] Other (please describe)


.github/workflows/build.yml
```diff
@@ -177,13 +177,22 @@ jobs:
...
         env:
+          DB_TEST_TRIGGER: ${{ contains(github.event.pull_request.labels.*.name, 'db') || contains(github.event.pull_request.labels.*.name, 'full build') }}
           GITHUB_EVENT_PULL_REQUEST_BASE_REF: ${{ github.event.pull_request.base.ref }}
           NEEDS_CHECK_SCIPY_TEST_TRIGGER_OUTPUTS_TEST_SCIPY: ${{ needs.check-scipy-test-trigger.outputs.test-scipy }}
         run: |
           export CHANGED_RECIPES=$(python ./tools/calc_diff.py \
             --base origin/${GITHUB_EVENT_PULL_REQUEST_BASE_REF} \
             --target ${{ github.sha }})
 
+          if [ "${DB_TEST_TRIGGER}" == "true" ]; then
```
hoodmane: This is getting pretty messy let's first move all this logic into Python.

.github/workflows/build.yml
```diff
@@ -402,6 +411,194 @@ jobs:
           path: test-results-${{ matrix.test-config.runtime }}.xml
         if: always()
 
+  test-db-integration:
```
hoodmane, line 406: This seems way too complicated, any way to test these with less than 180 lines of GitHub actions? I'm not sure I like the idea of maintaining that much yaml.

packages/mysqlclient/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - MySQLdb._mysql
+
+about:
+  home: https://github.com/PyMySQL/mysqlclient
+  PyPI: https://pypi.org/project/mysqlclient
+  summary: Python interface to MySQL backed by MariaDB Connector/C
+  license: GPL-2.0-or-later
+extra:
+  recipe-maintainers:
+    - ryanking13
```
hoodmane: Need to add end of file pre commit hook it seems.

packages/libmariadb/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+    mkdir -p \
+      ${INSTALL_DIR}/lib \
+      ${INSTALL_DIR}/include/mariadb/mysql \
+      ${INSTALL_DIR}/include/mariadb/mariadb
+
+    cp build/libmariadb/libmariadbclient.a ${INSTALL_DIR}/lib/libmariadb.a
+    ln -sf libmariadb.a ${INSTALL_DIR}/lib/libmariadbclient.a
+
+    cp \
```
hoodmane: It doesn't have `make install`? Could use a comment on why we can't use that. Also could this just be a wild card copy?

packages/libpq/meta.yaml
```diff
@@ -0,0 +1,109 @@
...
+    addlib ${INSTALL_DIR}/lib/libssl.a
+    addlib ${INSTALL_DIR}/lib/libcrypto.a
+    save
+    end
+    EOF
+    emar -M < libpq-selfcontained.mri
+    rm -f libpq-selfcontained.mri
+    emranlib ${INSTALL_DIR}/lib/libpq.a
+
+    cat > ${INSTALL_DIR}/bin/pg_config <<EOF
```
hoodmane: What's going on here? Isn't there a better way to do this?

packages/mysqlclient/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  sha256: 8ed20c5615a915da451bb308c7d0306648a4fd9a2809ba95c992690006306199
+
+requirements:
+  host:
+    - libopenssl
+    - libmariadb
+
+build:
+  script: |
+    export MYSQLCLIENT_CFLAGS="-I${WASM_LIBRARY_DIR}/include/mariadb -I${WASM_LIBRARY_DIR}/include"
```
hoodmane: Could these go in cflags / ldflags? If you can't use `WASM_LIBRARY_DIR` there maybe we should update pyodide-build.

packages/psycopg/test_psycopg_e2e.py
```diff
@@ -0,0 +1,149 @@
...
+                    )
+                    cursor.execute(
+                        sql.SQL("SELECT id, note, amount FROM {} WHERE id = %s").format(
+                            sql.Identifier(table_name)
+                        ),
+                        (record[0],),
+                    )
+                    assert cursor.fetchone() == record
+            finally:
+                with connection.cursor() as cursor:
```
hoodmane: What about a context manager `with drop_table(cursor, table_name)`? Also, does the cursor need to be closed and reopened?

.github/workflows/build.yml
ryanking13: Makes sense. Let me to that first.

ryanking13: Actually, this part seems to be not needed. Let's remove it for now.

packages/libpq/meta.yaml
ryanking13: Added a comment about it. pg_config is a binary so we cannot use a cross-compiled binary from other packages.

packages/libmariadb/meta.yaml
ryanking13: Changed to `make install`

.github/workflows/build.yml
ryanking13: Moved to a shell script.

packages/mysqlclient/meta.yaml
ryanking13: It uses pkg-config but the pkg-config was incorrect because we are building a static library. Add a patch in libmariadb instead.

packages/psycopg/test_psycopg_e2e.py
ryanking13: Sounds good. I tidied up a test code.

ryanking13:
Thanks for the review!

.github/workflows/build.yml
```diff
@@ -402,6 +403,87 @@ jobs:
...
+      - name: Install test dependencies
+        run: |
+          which python
+          echo y | python -m pip install -r requirements.txt
+
+      - name: Run DB integration tests
+        env:
+          MYSQL_SERVICE_ID: ${{ job.services.mysql.id }}
+          POSTGRES_SERVICE_ID: ${{ job.services.postgres.id }}
+          MYSQL_HOST: 127.0.0.1
```
hoodmane: What about moving these environment variables into the shell script too?

.github/workflows/build.yml
```diff
@@ -402,6 +403,87 @@ jobs:
...
+      - name: Download build artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: repodata
+          path: ./dist/
+
+      - name: Install test dependencies
+        run: |
+          which python
+          echo y | python -m pip install -r requirements.txt
```
hoodmane: Does `echo y` do anything? What does it do? Is there some flag to run in non interactive mode?

packages/libpq/meta.yaml
hoodmane: Is it possible to make it a separate file and put it in extras?

packages/libpq/meta.yaml
```diff
@@ -0,0 +1,113 @@
...
+      --without-ldap \
+      --without-libcurl \
+      --without-liburing \
+      --without-lz4 \
+      --without-readline \
+      --without-zlib \
+      --without-zstd
+
+    # force USE_OPENSSL=1
+    grep -q '#define USE_OPENSSL 1' src/include/pg_config.h
```
hoodmane: What does this do? Fail if it didn't set it?

packages/libpq/patches/0001-Disable-SO_PEERCRED-for-Emscripten.patch
```diff
@@ -0,0 +1,23 @@
+From: ryanking13 <def6488@gmail.com>
+Date: Thu, 13 Aug 2026 19:30:00 +0900
+Subject: [PATCH] Disable SO_PEERCRED for Emscripten
+
+Emscripten exposes the SO_PEERCRED constant but does not provide the struct
```
hoodmane, line 5: Is there anything to fix in emscripten here?

packages/psycopg-c/patches/0001-Use-Emscripten-endian-functions.patch
```diff
@@ -0,0 +1,22 @@
+From: ryanking13 <def6488@gmail.com>
+Date: Thu, 13 Aug 2026 18:30:00 +0900
+Subject: [PATCH] Use Emscripten endian functions
+
+Emscripten provides the standard endian conversion functions in endian.h,
```
hoodmane, line 5: Is there something we can fix here?

packages/mysqlclient/test_mysqlclient_e2e.py
```diff
@@ -0,0 +1,118 @@
...
+        "MYSQL_HOST",
+        "MYSQL_PORT",
+        "MYSQL_USER",
+        "MYSQL_PASSWORD",
+        "MYSQL_CA_FILE",
+    ]
+    missing = [name for name in required_names if not os.environ.get(name)]
+    if missing:
+        pytest.skip(
+            "Missing required MySQL test environment variables: " + ", ".join(missing)
```
hoodmane, line 26: Probably this should be an error?

packages/mysqlclient/test_mysqlclient_e2e.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    ca_file = Path(os.environ["MYSQL_CA_FILE"])
+    if not ca_file.is_file():
+        pytest.skip(f"MySQL CA file does not exist: {ca_file}")
```
hoodmane: Probably this should be an error? We should decide whether we want to run this via command line flags and then if it's missing what it needs it should be an error.

.github/workflows/build.yml
agriyakhetarpal: Yes, I don't think this is needed. `-y`/`--yes` is required for `pip uninstall`, not `pip install`.

packages/libopenssl/patches/0001-Allow-getentropy-for-Emscripten.patch
agriyakhetarpal, line 1: Note: since I previously removed `libopenssl` from pyodide-recipes as no recipe was using it, we should add it back, so that we remember to have these patches.

.github/workflows/build.yml
ryanking13: Looks like I added that in another job 2 years ago when adopting conda. I don't remember why I added that, but let me see if removing it works.

packages/psycopg-c/patches/0001-Use-Emscripten-endian-functions.patch
ryanking13: Probably we can upstream this to psycopg-c.

packages/libpq/meta.yaml
ryanking13: It was for double checking. Removed.

packages/libpq/patches/0001-Disable-SO_PEERCRED-for-Emscripten.patch
ryanking13: Yes possibly. Since Emscripten now has NODERAWSOCKETS thanks to Guy, making `getsockopt(SO_PEERCRED)` working should be able to fix this I guess.

Note that IIUC this function is used for unix socket, so I don't think this function will ever be used for us anyways.

packages/mysqlclient/test_mysqlclient_e2e.py
ryanking13: Right, changed to error.

.github/workflows/build.yml
ryanking13: Removed env variables other than `MYSQL_SERVICE_ID`, `POSTGRES_SERVICE_ID`. The leftovers are needed as it is a safe way to inject env variable in GHA (which I learned in  our security session).

hoodmane (approved):
Thanks!

PR #657, opened 2026-09-08
Remove cramjam and fastparquet

ariostas:
Cramjam and FastParquet haven't been built since #544.

I adapted the Cramjam CI to build Pyodide wheels in https://github.com/milesgranger/cramjam/pull/249, so there's no longer a need for a recipe here.

As for FastParquet, it was not being built due to the Cramjam dependency, but now they are retiring the project (see [their readme](https://github.com/dask/fastparquet/blob/main/README.rst)). So it's probably not worth it to keep the recipe here either.


ryanking13 (approved):
Thanks. It is good to know that cramjam has added PyEmscripten wheel in their CI!

PR #663, opened 2026-09-11
Remove pymupdf

ryanking13:
pymupdf has pyemscripten build in PyPI and the build is currently broken because of swig 4.5 release. I tried to fix that but it doesn't seem straightforward, so just removing it to unblock


ryanking13:
cc @linuxrider @sanzoghenzo PyMuPDF now has PyEmscripten build in its CI, so removing this recipe. Please forward any requests regarding PyMuPDF + Pyodide to upstream. Feel free to tag me if there are any blockers.

PR #665, opened 2026-09-14
Re-enable css-inline for PyEmscripten 2026

Stranger6667:
## Description

Re-enables the `css-inline` package. The latest PyPI release contains Pyodide-compatible wheels - not sure if this is still needed, but wanted to do it for completeness

## Type of change

- [ ] Adding a new package
- [ ] Bug fix
- [ ] Other (please describe)


ryanking13:
Thanks! I think we can cleanup these prebuilt packages before releasing Pyodide 315.

PR #666, opened 2026-09-14
Remove contourpy

ianthomas23:
ContourPy 1.4.0 uploaded pyemscripten wheels to [PyPI](https://pypi.org/project/contourpy/#files) for python 3.13, 3.14 and 3.15 and will continue to do so in future releases, so the recipe here can be removed.


agriyakhetarpal (approved):
Awesome, thanks @ianthomas23!

ryanking13:
Looks like we still needs this since we have matplotlib and bokeh recipe remaining in thie repository. Let's remove this altogether with matplotlib when matplotlib gets Pyodide ci.

agriyakhetarpal:
Oops, thanks! Sorry for the hasty merge; I thought that the test job was flaky and merged it 🤦🏻

PR #668, opened 2026-09-18
Remove mypy, librt, xxhash, and b2d recipes

ryanking13:
These packages have pyemscripten wheels on PyPI and IIRC we never had issues building these packages before.

`pyb2d is a package by @DerThorsten, but it has been archived since last year, so I don't think it has much value to keep it here.

cc: @ifduyue for viz (xxhash)


DerThorsten:
jeah it can be dropped

